Issue #89, opened 2014-03-07
Provide ability to provide custom scala.tools.nsc.Reporter in util-eval.


metasim:
It would be helpful to be able to specify the `scala.tools.nsc.Reporter` used by the private StringCompiler class for flexibility in error handling.

vkostyukov:
I belive this should be closed since it's fixed in #90.

PR #90, opened 2014-03-13
Added ability to override `Reporter` used during compilation. 

metasim:
See issue #89.


mosesn:
@metasim sorry this fell through the cracks.  I just merged it in internally, and it should appear on develop in the next couple of weeks.  Normally it would be weekly, but we have a problem in our system for pushing to github right now.

mosesn:
This is finally in develop, sorry for the long, long delay.  https://github.com/twitter/util/commit/963de9f46f1f76502bc3bd79398854003aff085e

metasim:
Thanks!

PR #108, opened 2014-08-21
Proposal to allow eval to be configured with a custom CP

andypetrella:
Problem:
For whatever reason the classpath from this.getClass.getClassLoader was quite empty during the test-run within maven ( only the jdk jars + surefire + jacoco ones were found o.O ).
It might be due to the fact that things are launched in a fork thread (ForkJoinPool used by test AFAICT).
So when I was using custom import (even scala of course), classes were not found and thus the file failed to compile.

Solution:
Added an extra parameter to Eval's constructor that takes a function producing a `List[List[String]]` (keeping the signature of `getClasspath`). 
Two implementations provided:
* the original policy using the classpath accessible relatively to the Eval.class
* another one using the sys property java.class.path to use the whole classpath
The no-parameter constructor has been updated to use the original strategy for backward-compat.

Result:
Now, we can use the full classpath detected by the jvm, however, using this way we loose the "executable-jar" functionality when the direct classloader has only one jar.

selfie-7

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -2,7 +2,7 @@ import sbt.Keys._
 import sbt._
 
 object Util extends Build {
-  val libVersion = "6.19.0"
+  val libVersion = "6.19.1-SNAPSHOT"
   val zkVersion = "3.3.4"
   val zkDependency = "org.apache.zookeeper" % "zookeeper" % zkVersion excludeAll(
     ExclusionRule("com.sun.jdmk", "jmxtools"),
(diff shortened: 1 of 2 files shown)


mosesn:
@andypetrella Did you want to continue with this PR?  We changed our minds about the 2.11 thing, by the way, since it turned out it was much easier to support than we expected.

koertkuipers:
@andypetrella 
what do you mean we lose the "executable-jar functionality when the direct classloader has only one jar."
trying to understand the consequences....
best, koert

andypetrella:
@mosesn indeed, why not. Actually, I stopped using it when you said that it'll be discontinued :-D
@koertkuipers *shame* I just CANNOT remember what I had in mind/figured out -- it's been awhile since then, I probably wiped out my brain...

I guess that giving it another try will be necessary.

I'll try to find some time to process the comments if you'd like to. Just tell me ;-).

cheers,
andy

mosesn:
@andypetrella probably wise anyway, it's sort of difficult to test, and we're discouraging folks from using it internally.

PR #109, opened 2014-09-04
Allow manual flushing of a batcher with flushBatch method

iceberg901:
Implements part of a TODO item listed in comments in BatchExecutor.scala

Future.batched now returns an instance of class Batcher, which has the added method flushBatch to allow manual flushing/execution of remaining items in the batcher.

### DIFF
--- a/util-core/src/main/scala/com/twitter/util/BatchExecutor.scala
+++ b/util-core/src/main/scala/com/twitter/util/BatchExecutor.scala
@@ -98,6 +98,14 @@ private[util] class BatchExecutor[In, Out](
     promise
   }
 
+  def flushNow(): Unit = {
+    val doAfter = synchronized {
+      flushBatch()
+    }
+
+    doAfter()
+  }
+
   def scheduleFlushIfNecessary() {
     if (timeThreshold < Duration.Top && scheduled.isEmpty)
       scheduled = Some(new ScheduledFlush(timeThreshold, timer))
--- a/util-core/src/main/scala/com/twitter/util/BatchExecutor.scala
+++ b/util-core/src/main/scala/com/twitter/util/BatchExecutor.scala
@@ -98,6 +98,7 @@ private[util] class BatchExecutor[In, Out](
     promise
   }
 
+  /** Immediately processes all unprocessed requests */
   def flushNow(): Unit = {
     val doAfter = synchronized {
       flushBatch()
--- a/util-core/src/main/scala/com/twitter/util/BatchExecutor.scala
+++ b/util-core/src/main/scala/com/twitter/util/BatchExecutor.scala
@@ -22,9 +22,8 @@ import scala.collection.mutable
  * - Rather than having separate sizeThreshold and sizePercentile parameters,
  *   just have a single call-by-name sizeThreshold parameter, and let the
  *   caller implement whatever logic they want to compute the next batch size.
- * - Return an instance of a new class (Batcher?) which extends
- *   `In => Future[Out]`. Could support things like querying the queue size,
- *   forcing a flush, attaching callbacks to flush operations, etc.
+ * - Add more functionality to class Batcher. Could support things like querying the queue size,
+ *   attaching callbacks to flush operations, etc.
  */
 private[util] class BatchExecutor[In, Out](
   sizeThreshold: Int,
--- a/util-core/src/main/scala/com/twitter/util/Batcher.scala
+++ b/util-core/src/main/scala/com/twitter/util/Batcher.scala
@@ -6,10 +6,11 @@ class Batcher[In, Out] private[util](
 )(
   implicit timer: Timer
 ) extends Function1[In, Future[Out]] { batcher =>
-  /** Enqueues requests for a batched [[com.twitter.util.Future]]
-    *
-    * Processes all enqueued requests after the batch size is reached
-    */
+  /**
+   * Enqueues requests for a batched [[com.twitter.util.Future]]
+   *
+   * Processes all enqueued requests after the batch size is reached
+   */
   def apply(t: In): Future[Out] = executor.enqueue(t)
 
   /** Immediately processes all unprocessed requests */
--- a/util-core/src/main/scala/com/twitter/util/Batcher.scala
+++ b/util-core/src/main/scala/com/twitter/util/Batcher.scala
@@ -1,27 +1,17 @@
 package com.twitter.util
 
-import java.util.logging.{Level, Logger}
-
-import scala.collection.mutable
-
-/** Provides a clean, lightweight interface for controlling a BatchExecutor
-  *
-  * @constructor create a new Batcher for a BatchExecutor
-  * @param executor the BatchExector to be used
-  */
-class Batcher[In, Out](
+/** Provides a clean, lightweight interface for working with batched Futures */
+private[util] class Batcher[In, Out](
   executor: BatchExecutor[In, Out]
 )(
   implicit timer: Timer
 ) extends Function1[In, Future[Out]] { batcher =>
-  import java.util.logging.Level.WARNING
-
-  /** Enqueues requests for the BatchExecutor */
+  /** Enqueues requests for a batched Future */
   def apply(t: In): Future[Out] = executor.enqueue(t)
 
   /** Immediately processes all unprocessed requests */
   def flushBatch(): Unit = {
-    val doAfter = synchronized {
+    val doAfter = executor.synchronized {
       executor.flushBatch()
     }
 
--- a/util-core/src/main/scala/com/twitter/util/Batcher.scala
+++ b/util-core/src/main/scala/com/twitter/util/Batcher.scala
@@ -1,0 +1,28 @@
+package com.twitter.util
+
+import java.util.logging.{Level, Logger}
+
+import scala.collection.mutable
+
+class Batcher[In, Out](
+  sizeThreshold: Int,
+  timeThreshold: Duration = Duration.Top,
+  sizePercentile: => Float = 1.0f,
+  f: Seq[In] => Future[Seq[Out]]
+)(
+  implicit timer: Timer
+) extends Function1[In, Future[Out]] { batcher =>
+  import java.util.logging.Level.WARNING
+
+  val executor = new BatchExecutor[In, Out](sizeThreshold, timeThreshold, sizePercentile, f)
+
+  def apply(t: In): Future[Out] = executor.enqueue(t)
+
+  def flushBatch() = {
+    val doAfter = synchronized {
+      executor.flushBatch()
+    }
+
+    doAfter()
+  }
+}
--- a/util-core/src/main/scala/com/twitter/util/Batcher.scala
+++ b/util-core/src/main/scala/com/twitter/util/Batcher.scala
@@ -4,21 +4,23 @@ import java.util.logging.{Level, Logger}
 
 import scala.collection.mutable
 
+/** Provides a clean, lightweight interface for controlling a BatchExecutor
+  *
+  * @constructor create a new Batcher for a BatchExecutor
+  * @param executor the BatchExector to be used
+  */
 class Batcher[In, Out](
-  sizeThreshold: Int,
-  timeThreshold: Duration = Duration.Top,
-  sizePercentile: => Float = 1.0f,
-  f: Seq[In] => Future[Seq[Out]]
+  executor: BatchExecutor[In, Out]
 )(
   implicit timer: Timer
 ) extends Function1[In, Future[Out]] { batcher =>
   import java.util.logging.Level.WARNING
 
-  val executor = new BatchExecutor[In, Out](sizeThreshold, timeThreshold, sizePercentile, f)
-
+  /** Enqueues requests for the BatchExecutor */
   def apply(t: In): Future[Out] = executor.enqueue(t)
 
-  def flushBatch() = {
+  /** Immediately processes all unprocessed requests */
+  def flushBatch(): Unit = {
     val doAfter = synchronized {
       executor.flushBatch()
     }
--- a/util-core/src/main/scala/com/twitter/util/Batcher.scala
+++ b/util-core/src/main/scala/com/twitter/util/Batcher.scala
@@ -1,11 +1,11 @@
 package com.twitter.util
 
-/** Provides a clean, lightweight interface for working with a batched [[com.twitter.util.Future]] */
+/** Provides an interface for working with a batched [[com.twitter.util.Future]] */
 class Batcher[In, Out] private[util](
   executor: BatchExecutor[In, Out]
 )(
   implicit timer: Timer
-) extends Function1[In, Future[Out]] { batcher =>
+) extends (In => Future[Out]) { batcher =>
   /**
    * Enqueues requests for a batched [[com.twitter.util.Future]]
    *
--- a/util-core/src/main/scala/com/twitter/util/Batcher.scala
+++ b/util-core/src/main/scala/com/twitter/util/Batcher.scala
@@ -1,14 +1,14 @@
 package com.twitter.util
 
-/** Provides a clean, lightweight interface for working with batched Futures */
+/** Provides a clean, lightweight interface for working with a batched [[com.twitter.util.Future]] */
 class Batcher[In, Out] private[util](
   executor: BatchExecutor[In, Out]
 )(
   implicit timer: Timer
 ) extends Function1[In, Future[Out]] { batcher =>
-  /** Enqueues requests for a batched Future
+  /** Enqueues requests for a batched [[com.twitter.util.Future]]
     *
-    * When the batch size is reached, all enqueued requests are processed
+    * Processes all enqueued requests after the batch size is reached
     */
   def apply(t: In): Future[Out] = executor.enqueue(t)
 
--- a/util-core/src/main/scala/com/twitter/util/Batcher.scala
+++ b/util-core/src/main/scala/com/twitter/util/Batcher.scala
@@ -14,7 +14,5 @@ class Batcher[In, Out] private[util](
   def apply(t: In): Future[Out] = executor.enqueue(t)
 
   /** Immediately processes all unprocessed requests */
-  def flushBatch(): Unit = {
-    executor.flushNow()
-  }
+  def flushBatch(): Unit = executor.flushNow()
 }
--- a/util-core/src/main/scala/com/twitter/util/Batcher.scala
+++ b/util-core/src/main/scala/com/twitter/util/Batcher.scala
@@ -1,12 +1,15 @@
 package com.twitter.util
 
 /** Provides a clean, lightweight interface for working with batched Futures */
-private[util] class Batcher[In, Out](
+class Batcher[In, Out] private[util](
   executor: BatchExecutor[In, Out]
 )(
   implicit timer: Timer
 ) extends Function1[In, Future[Out]] { batcher =>
-  /** Enqueues requests for a batched Future */
+  /** Enqueues requests for a batched Future
+    *
+    * When the batch size is reached, all enqueued requests are processed
+    */
   def apply(t: In): Future[Out] = executor.enqueue(t)
 
   /** Immediately processes all unprocessed requests */
--- a/util-core/src/main/scala/com/twitter/util/Batcher.scala
+++ b/util-core/src/main/scala/com/twitter/util/Batcher.scala
@@ -15,10 +15,6 @@ class Batcher[In, Out] private[util](
 
   /** Immediately processes all unprocessed requests */
   def flushBatch(): Unit = {
-    val doAfter = executor.synchronized {
-      executor.flushBatch()
-    }
-
-    doAfter()
+    executor.flushNow()
   }
 }
--- a/util-core/src/main/scala/com/twitter/util/Future.scala
+++ b/util-core/src/main/scala/com/twitter/util/Future.scala
@@ -533,6 +533,10 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
    *       ...
    *     }
    *
+   * To force the batcher to immediately process all unprocessed requests:
+   *
+   *     batcher.flushBatch
+   *
    * A batcher's size can be controlled at runtime with the `sizePercentile`
    * function argument. This function returns a float between 0.0 and 1.0,
    * representing the fractional size of the `sizeThreshold` that should be
@@ -546,8 +550,8 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
     f: Seq[In] => Future[Seq[Out]]
   )(
     implicit timer: Timer
-  ): In => Future[Out] = {
-    new BatchExecutor[In, Out](sizeThreshold, timeThreshold, sizePercentile, f)
+  ): Batcher[In, Out] = {
+    new Batcher[In, Out](sizeThreshold, timeThreshold, sizePercentile, f)
   }
 }
 
--- a/util-core/src/main/scala/com/twitter/util/Future.scala
+++ b/util-core/src/main/scala/com/twitter/util/Future.scala
@@ -535,7 +535,7 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
    *
    * To force the batcher to immediately process all unprocessed requests:
    *
-   *     batcher.flushBatch
+   *     batcher.flushBatch()
    *
    * A batcher's size can be controlled at runtime with the `sizePercentile`
    * function argument. This function returns a float between 0.0 and 1.0,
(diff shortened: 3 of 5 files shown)


mosesn:
@iceberg901 do you want to continue with this PR?

iceberg901:
@mosesn Yes I do, I'll start working on addressing your comments.

mosesn:
Awesome, let me know if you need any help.

util-core/src/main/scala/com/twitter/util/Future.scala
```diff
@@ -514,8 +518,8 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
     f: Seq[In] => Future[Seq[Out]]
   )(
     implicit timer: Timer
-  ): In => Future[Out] = {
-    new BatchExecutor[In, Out](sizeThreshold, timeThreshold, sizePercentile, f)
+  ): Batcher[In, Out] = {
```
iceberg901: So how do you want to proceed here?  I don't have any insight into how often you can or want to do major releases.  I'm happy to add another endpoint if that helps this get through faster, but overall I think returning a more full-featured object is the way to go in the end.

iceberg901:
@mosesn No, I don't think we want to return BatchExecutor.  Returning a wrapper of some kind gives us the luxury of providing a clean, simple interface not clouded by scary implementation details.  I assume that's why the original implementer(s) didn't return BatchExecutor in the first place.  However, I think the current interface doesn't provide enough functionality, so I think the right answer is to return Batcher and to keep augmenting it if ever we want to expose additional functionality.

util-core/src/main/scala/com/twitter/util/Future.scala
```diff
@@ -501,6 +501,10 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
    *       ...
    *     }
    *
+   * To force the batcher to immediately process all unprocessed requests:
+   *
+   *     batcher.flushBatch
```
iceberg901: Done

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,28 @@
+package com.twitter.util
+
+import java.util.logging.{Level, Logger}
+
+import scala.collection.mutable
+
+class Batcher[In, Out](
```
iceberg901: Added

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,28 @@
...
+  sizeThreshold: Int,
```
iceberg901: Good idea.  Done.

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,28 @@
...
+)(
+  implicit timer: Timer
+) extends Function1[In, Future[Out]] { batcher =>
+  import java.util.logging.Level.WARNING
+
+  val executor = new BatchExecutor[In, Out](sizeThreshold, timeThreshold, sizePercentile, f)
+
+  def apply(t: In): Future[Out] = executor.enqueue(t)
+
+  def flushBatch() = {
```
iceberg901: Done

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,28 @@
...
+    val doAfter = synchronized {
```
iceberg901: I wrote some more tests to verify these scenarios and make sure that regular execution, manual flushes, and timed flushes don't interfere with each other.  Please let me know if you think the test coverage is sufficient.

iceberg901:
Ok, I've addressed all your comments except the one about not returning a new type from Future.batched.  Please let me know how you want to proceed there.  And give me feedback on the tests please.  Finally, all of my tests are green but there seems to be some problems with your test setup in general - red tests, exceptions getting thrown, etc.  Should I worry about this or does it all get take care of when you integrate?

mosesn:
Don't worry about the tests, it's something wrong with a new feature of travis-ci that we're trying out.

I agree with you that exposing BatchExecutor directly would be a mistake, but changing the return type of Future#batched also definitely breaks binary compatibility.  With that said, we've relaxed how we feel about breaking binary compatibility in the last few months, so we can try merging it in and see how difficult it is.

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,30 @@
+package com.twitter.util
+
+import java.util.logging.{Level, Logger}
+
+import scala.collection.mutable
+
+/** Provides a clean, lightweight interface for controlling a BatchExecutor
```
mosesn: This should be formatted like so:

```scala
/**
  * Provides ....
  * ...
  */
```

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,30 @@
...
+  *
+  * @constructor create a new Batcher for a BatchExecutor
```
mosesn: rm the `@constructor` line, doesn't really add anything

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,30 @@
...
+  * @param executor the BatchExector to be used
```
mosesn: rm this one too.  it's pretty obvious.

util-core/src/main/scala/com/twitter/util/Batcher.scala
mosesn: As you mentioned before, `BatchExecutor` is an internal implementation detail, so when giving folks instructions in how to use `Batcher`, we shouldn't ask them to understand `BatchExecutor`.

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,30 @@
...
+  */
+class Batcher[In, Out](
```
mosesn: We should make the constructor private[util] since folks outside of util won't have access to a `BatchExecutor` anyway.

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,30 @@
...
+  implicit timer: Timer
+) extends Function1[In, Future[Out]] { batcher =>
+  import java.util.logging.Level.WARNING
+
+  /** Enqueues requests for the BatchExecutor */
+  def apply(t: In): Future[Out] = executor.enqueue(t)
+
+  /** Immediately processes all unprocessed requests */
+  def flushBatch(): Unit = {
+    val doAfter = synchronized {
```
mosesn: this must be synchronized on the `BatchExecutor`.

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,30 @@
...
+  * @param executor the BatchExector to be used
+  */
+class Batcher[In, Out](
+  executor: BatchExecutor[In, Out]
+)(
+  implicit timer: Timer
+) extends Function1[In, Future[Out]] { batcher =>
+  import java.util.logging.Level.WARNING
+
+  /** Enqueues requests for the BatchExecutor */
```
mosesn: don't reveal implementation details

util-core/src/main/scala/com/twitter/util/Batcher.scala
mosesn: Testing for concurrency is very difficult, it's easier to just reason about the java memory model.  If we start running into big problems, we can go write some multithreaded tests to try to trigger the race condition, but it's fine for now to stick with singlethreaded tests.

With that said, the guarantee that we want is that there is only one caller of flushBatch() at a time.  The way `BatchExecutor` achieves this is by synchronizing on itself, which creates a happens-before relationship between any two calls to flushBatch().  I think the only way for us to do the same thing is to synchronize on the same thing (ie the `BatchExecutor`).

util-core/src/test/scala/com/twitter/util/FutureTest.scala
```diff
@@ -308,6 +308,65 @@ class FutureTest extends WordSpec with MockitoSugar with GeneratorDrivenProperty
...
         }
 
+        "execute when flushBatch is called" in {
+          val f = mock[Seq[Int] => Future[Seq[Int]]]
+          val batcher = Future.batched(4)(f)
+
+          batcher(1)
+          batcher(2)
+          batcher(3)
+          batcher.flushBatch
```
mosesn: add parens

util-core/src/test/scala/com/twitter/util/FutureTest.scala
```diff
@@ -308,6 +308,65 @@ class FutureTest extends WordSpec with MockitoSugar with GeneratorDrivenProperty
...
+        "only execute for remaining items when flushBatch is called after size threshold is reached" in {
+          val f = mock[Seq[Int] => Future[Seq[Int]]]
+          val batcher = Future.batched(4)(f)
+
+          batcher(1)
+          batcher(2)
+          batcher(3)
+          batcher(4)
+          batcher(5)
+          batcher.flushBatch
```
mosesn: add parens

util-core/src/main/scala/com/twitter/util/Batcher.scala
iceberg901: So passing the BatchExecutor as a constructor parameter was your idea to reduce the clutter of having to pass all of the BatchExecutor's constructor params through two consecutive constructor methods.  And no one's creating Batchers directly, they're getting them from Future.batched.

So I guess I'm wondering what you think now: should we back off having the BatchExecutor as a parameter to the Batcher constructor, or should we just make the constructor private[util] and leave it at that.

util-core/src/test/scala/com/twitter/util/FutureTest.scala
```diff
@@ -308,6 +308,65 @@ class FutureTest extends WordSpec with MockitoSugar with GeneratorDrivenProperty
...
+
+        "only execute once when time threshold is reached after flushBatch is called" in {
+          val f = mock[Seq[Int] => Future[Seq[Int]]]
+          val batcher = Future.batched(4, 3.seconds)(f)
+
+          Time.withCurrentTimeFrozen { control =>
+            batcher(1)
+            batcher(2)
+            batcher(3)
+            batcher.flushBatch
```
mosesn: add parens

util-core/src/test/scala/com/twitter/util/FutureTest.scala
```diff
@@ -308,6 +308,65 @@ class FutureTest extends WordSpec with MockitoSugar with GeneratorDrivenProperty
...
+          val f = mock[Seq[Int] => Future[Seq[Int]]]
+          val batcher = Future.batched(4, 3.seconds)(f)
+
+          Time.withCurrentTimeFrozen { control =>
+            batcher(1)
+            batcher(2)
+            batcher(3)
+            control.advance(10.seconds)
+            timer.tick()
+            batcher.flushBatch
```
mosesn: add parens

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,30 @@
...
+  *
+  * @constructor create a new Batcher for a BatchExecutor
+  * @param executor the BatchExector to be used
+  */
+class Batcher[In, Out](
+  executor: BatchExecutor[In, Out]
+)(
+  implicit timer: Timer
+) extends Function1[In, Future[Out]] { batcher =>
+  import java.util.logging.Level.WARNING
```
mosesn: unused

util-core/src/main/scala/com/twitter/util/BatchExecutor.scala
```diff
@@ -22,9 +22,8 @@ import scala.collection.mutable
  * - Rather than having separate sizeThreshold and sizePercentile parameters,
  *   just have a single call-by-name sizeThreshold parameter, and let the
  *   caller implement whatever logic they want to compute the next batch size.
- * - Return an instance of a new class (Batcher?) which extends
- *   `In => Future[Out]`. Could support things like querying the queue size,
- *   forcing a flush, attaching callbacks to flush operations, etc.
+ * - Add more functionality to class Batcher. Could support things like querying the queue size,
```
mosesn: this doesn't really belong in scaladoc.  Could you move it out into a `// TODO` comment?

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,30 @@
+package com.twitter.util
+
+import java.util.logging.{Level, Logger}
```
mosesn: unused, as far as I can tell.

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,30 @@
...
+
+import scala.collection.mutable
```
mosesn: unused, as far as I can tell.

mosesn:
This looks good–there's one synchronization thing, and a few nits about docs, and a few other nitpicks, and then we should be ready to rock :sunrise:

util-core/src/main/scala/com/twitter/util/Batcher.scala
mosesn: No, I think BatchExecutor as a parameter to the Batcher constructor is still a good idea, but Future.batched should be the only way to get a Batcher.  Let's make the constructor private[util]. :ship:

util-core/src/main/scala/com/twitter/util/BatchExecutor.scala
iceberg901: Future.scala has TODO: inside of scaladoc for its methods all over the place.  Maybe we should leave this this way for consistency?  If not, where in the file would you like to see it?  And formatted like this?:

```scala
// TODO: Possible future improvements:
// - Rather than having separate sizeThreshold and sizePercentile parameters,
//   just have a single call-by-name sizeThreshold parameter, and let the
//   caller implement whatever logic they want to compute the next batch size.
// - Add more functionality to class Batcher. Could support things like querying the queue size,
//   attaching callbacks to flush operations, etc.
```

mosesn: The only TODOs I could find in Future.scala were about future deprecation, which is actually probably of interest to the end-user.  My general philosophy is that TODOs that are about changing semantics are for librarians, so they should be in normal comments.

That formatting is fine, I think just sticking it in the line beneath the constructor for BatchExecutor is good.

EDIT: actually, I just remembered that BatchExecutor is private, so it won't get published as scaladoc anyway.  I don't feel as strongly now, so feel free to ignore this.

iceberg901:
Ok, I think I took care of everything.  Let me know if anything was missed.  What happens next?

Do you need me to squash the commits?  Use some specific style for the commit message?  Anything else?  Let me know.

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,20 @@
+package com.twitter.util
+
+/** Provides a clean, lightweight interface for working with batched Futures */
+private[util] class Batcher[In, Out](
```
mosesn: this class can't be private if we're exposing it as a return type for Future.batched.  can you just make the constructor private instead?  That would look like:

```scala
class Batcher[In, Out] private[util](...
```

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,20 @@
...
+  executor: BatchExecutor[In, Out]
+)(
+  implicit timer: Timer
+) extends Function1[In, Future[Out]] { batcher =>
+  /** Enqueues requests for a batched Future */
```
mosesn: feel free to ignore:
Maybe also note that if apply makes it hit the batch size, it will automatically process the batch.

s/Future/[[com.twitter.util.Future]]

mosesn:
Just two more things.  After that, I'll have a couple more folks take a look at it, and we'll take it from there.  You shouldn't have to do a thing.

util-core/src/main/scala/com/twitter/util/Batcher.scala
iceberg901: OK thanks will do.  Scala syntax often has me confused about where the physical delineation is in the code between the class and its constructor :)

util-core/src/main/scala/com/twitter/util/Batcher.scala
iceberg901: Done - comment look good to you?

iceberg901: Also, you are speaking to me in regex here I guess?  You want me to fully qualify all my references to com.twitter.util.Future?  Or just the one on this line?  And only in comments or in comments and code?

mosesn: More I'm saying that if we wrap it in double square brackets then scaladoc will render it as an internal link.  It's not a huge deal, but it can be nice.  I'm OK with merging as-is.

mosesn:
LGTM, thanks for bearing with me!  I'll try to get more eyes on this and then we can merge it in.

iceberg901:
Thanks!  Nice combing through the details with you :)

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,24 @@
+package com.twitter.util
+
+/** Provides a clean, lightweight interface for working with a batched [[com.twitter.util.Future]] */
+class Batcher[In, Out] private[util](
+  executor: BatchExecutor[In, Out]
+)(
+  implicit timer: Timer
+) extends Function1[In, Future[Out]] { batcher =>
```
vkostyukov: Can we extend `In => Future[Out]` instead?

util-core/src/main/scala/com/twitter/util/Future.scala
stevegury: We will run some test internally to see if it breaks anything.
If it doesn't, I think it's fine to take the binary incompatible change, just update the CHANGES file and explain the change.
This class is not widely used enough to require to wait for a major release (we decided not to strictly following semantic versioning).

vkostyukov:
Looks good to me! @iceberg901 do you mind to squash the commits of this PR into one? You can use `git rebase -i` and then `git --force push origin your-branch`.

mosesn:
@vkostyukov git review submit squashes the changes for us, so I don't think it should matter?

vkostyukov:
@mosesn cool! Nevermind then.

util-core/src/main/scala/com/twitter/util/Batcher.scala
iceberg901: I don't have a strong feeling about this.  I think the reason I went with Function1 was that I was having trouble getting it to compile the other way.  I've since found that I can get it to work only with parens:

```scala
) extends (In => Future[Out]) { batcher =>
```

@mosesn @vkostyukov please let me know if you want me to make this change

iceberg901:
Please let me know if you want me to make the change on Batcher above.  I take it I don't need to squash the commits?  Anything else I need to do?

mosesn:
Could you make the change to the ChangeLog that Steve mentioned?

iceberg901:
Sure, would you put it under New Features or API Changes?  Or do I describe the flushBatch method under New Features and note the change of return type of Future.batched under API Changes?

mosesn:
Yeah, I think both is a good idea.

util-core/src/test/scala/com/twitter/util/FutureTest.scala
```diff
@@ -308,6 +308,65 @@ class FutureTest extends WordSpec with MockitoSugar with GeneratorDrivenProperty
...
+          val batcher = Future.batched(4)(f)
+
+          batcher(1)
+          batcher(2)
+          batcher(3)
+          batcher(4)
+          batcher(5)
+          batcher.flushBatch()
+
+          verify(f, times(1)).apply(Seq(1,2,3,4))
```
dschobel: can you move the first verify (ln 334) above the flushBatch call? that would make the behavior clearer because as-is it looks like neither batch is dispatched until flushing.

iceberg901: done

iceberg901:
@mosesn Changelog updated, what else?

dschobel:
this LGTM

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,24 @@
...
+  /**
+   * Enqueues requests for a batched [[com.twitter.util.Future]]
+   *
+   * Processes all enqueued requests after the batch size is reached
+   */
+  def apply(t: In): Future[Out] = executor.enqueue(t)
+
+  /** Immediately processes all unprocessed requests */
+  def flushBatch(): Unit = {
```
kevinoliver: if there are guarantees regarding ordering and such, they should be documented. its not obvious from reading the diff why you synchronize on executor here.

and if there are guarantees, what prevents others callers from using the executor incorrectly?

util-core/src/main/scala/com/twitter/util/Batcher.scala
```diff
@@ -0,0 +1,24 @@
+package com.twitter.util
+
+/** Provides a clean, lightweight interface for working with a batched [[com.twitter.util.Future]] */
```
kevinoliver: please omit "clean, lightweight" — I don't think they add any extra value here.

util-core/src/main/scala/com/twitter/util/Batcher.scala
kevinoliver: i think its more idiomatic as @vkostyukov wrote it.

util-core/src/main/scala/com/twitter/util/Batcher.scala
iceberg901: Done

util-core/src/main/scala/com/twitter/util/Batcher.scala
iceberg901: Ok two votes for ```In => Future[Out]```, so I changed it.

util-core/src/main/scala/com/twitter/util/Batcher.scala
iceberg901: * I did not implement the BatchExecutor, but I make no assumptions about order of execution of the individual requests within a batch.  Since there is a one-to-one correlation between the requests I submit and the Futures returned, I as a client can at least decide what order I want to process the results in, regardless of what order they complete.

* As far as why I synchronize, the inline documentation for BatchExecutor tells me I must do so when calling .flushBatch() (BatchExecutor.scala line 107).

* Finally, how do we prevent other callers from using the executor incorrectly: by making BatchExecutor private\[util\] (this was done already by whoever implemented BatchExecutor) and only providing access to it through the Batcher interface (which is what I am adding).

If there is some part of what I've just explained here that you would like me to document in the code, please let me know.

kevinoliver: thanks for the explanation, i'm new to BatchExecutor — that's a funky interface.

i think a comment in your code as to why you synchronize there would be quite helpful for future maintainers.

iceberg901: Here's how I decided to address this:

I moved the synchronization inside a new method in the BatchExecutor called .flushNow().  So, Batcher.flushBatch() just calls BatchExecutor.flushNow()

This is good IMO because:
* A person new to the code will find all synchronization logic inside of one file, BatchExecutor.scala, and they will easily be able to see that all calls to BatchExecutor.flushBatch are wrapped in synchronize blocks.
* If that person still doesn't understand what's going on, the comments explaining that synchronization is required are in the same file, not a a different file.
* Batcher.scala stays simple

Does this address your concerns?

mosesn:
LGTM

kevinoliver:
lgtm too, thanks for your patience working with us. i'll try to get this patch rolling internally today.

kevinoliver:
Ok, this just got merged locally and will show up on the develop branch soon, once we get some issues on our side sorted out. Thanks again for the patch and your patience.

iceberg901:
Great thanks!  For future reference, should I open PRs against the develop branch?

mosesn:
Yes, but this PR predates the `develop` branch so no worries! :+1:

mosesn:
:ice_cream: this hit develop woo https://github.com/twitter/util/commit/616479c2acff63f3b93c8be8895ef285096a07f6

iceberg901:
Awesome, thanks!  Do you have a general timetable for your next release?

mosesn:
Not yet, but we'll send out an email on the finaglers listserv when we do it.

PR #121, opened 2015-01-10
Unit test breaks in other locales

eirslett:
Some locales represent floats as "12,3" instead of "12.3", so StorageUnitTest fails.
This PR forces English locale to be used when formatting storage units.
It changes code behaviour (rather than the test), but I guess it shouldn't matter that much.


mosesn:
LGTM

stevegury:
LGTM, I'm pulling this internally.

mosesn:
:boom:  merged in https://github.com/twitter/util/commit/0b73bbc3adc4e1766e6e3d79f233ea29f7a240c5

PR #122, opened 2015-01-11
Add helpers for Java 8 lambda integration

eirslett:
I tried to overload the "map", "flatMap" etc. methods in c.t.u.Future directly, so they could accept Java lambdas without using a converter function. But I couldn't make it compile. (Does anybody know how it can be done?)

This commit lets you write Java code that looks like this:

```java
Future.value(3)
  .flatMap(func(i -> Future.value(2*i)))
  .map(func(i -> i + 3))
  .onSuccess(cons(System.out::println));
```

or

```java
userService.getUser("john")
  .flatMap(func(u -> profileService.getProfile(u.id)))
  .map(func(Templates::renderProfile))
  .onSuccess(cons(System.out::println));
```

**Commit message:**
> The method c.t.u.Function.func converts a Java 8 lambda
> into a c.t.u.Function, which can be used in Twitter's
> functional APIs (e.g. Futures)
> 
> example:
>
>```java
> import static com.twitter.util.Function.func;
> Function<String, String> fun = func(String::toUpperCase);
> ```
>
> There is a method c.t.u.Function.exfunc that also
> deals with checked exceptions.
> 
> You often need consumers, functions of type T => Unit,
> e.g. for future.onSuccess. In these cases, you can use
> Function.cons to convert a lambda-style consumer into
> a Scala function
> 
> example:
> ```java
> Function<String, BoxedUnit> fun = cons(System.out::println);
> ```
> 
> Later, Scala 2.12 will have native support for Java 8 lambdas.

### DIFF
--- a/util-core/src/main/java/com/twitter/util/ExceptionalJavaConsumer.java
+++ b/util-core/src/main/java/com/twitter/util/ExceptionalJavaConsumer.java
@@ -1,0 +1,10 @@
+package com.twitter.util;
+
+/**
+ * Used for Java 8 interop.
+ * This is a SAM version of c.t.u.ExceptionalFunction (returning Unit/void).
+ */
+// @FunctionalInterface
+public interface ExceptionalJavaConsumer<T> {
+    public void apply(T value) throws Throwable;
+}
--- a/util-core/src/main/scala/com/twitter/util/Function.scala
+++ b/util-core/src/main/scala/com/twitter/util/Function.scala
@@ -47,6 +47,58 @@ object Function {
    * Creates `() => A` function from given `Callable`.
    */
   def ofCallable[A](c: Callable[A]): () => A = () => c.call()
+
+  /**
+   * Creates a T => R from a JavaFunction. Used for easier interop
+   * between Java 8 and Twitter Util libraries.
+   */
+  def function[T, R](f: JavaFunction[T, R]) = new Function[T, R] {
+    override def apply(value: T): R = f(value)
+  }
+
+  /**
+   * Short form of `function`
+   */
+  def f[T, R](f: JavaFunction[T, R]) : Function[T, R] = function(f)
+
+  /**
+   * Creates a T => Unit from a JavaConsumer.
+   * Useful for e.g. future.onSuccess
+   */
+  def consumer[T](f: JavaConsumer[T]) = new Function[T, Unit] {
+    override def apply(value: T): Unit = f(value)
+  }
+
+  /**
+   * Short form of `consumer`
+   */
+  def c[T](f: JavaConsumer[T]) : Function[T, Unit] = consumer(f)
+
+  /**
+   * like `function`, but deals with checked exceptions as well
+   */
+  def exceptionalFunction[T, R](f: ExceptionalJavaFunction[T, R]) = new ExceptionalFunction[T, R] {
+    @throws(classOf[Throwable])
+    override def applyE(value: T): R = f(value)
+  }
+
+  /**
+   * Short form of `exceptionalFunction`
+   */
+  def xf[T, R](f: ExceptionalJavaFunction[T, R]) : ExceptionalFunction[T, R] = exceptionalFunction(f)
+
+  /**
+   * like `consumer`, but deals with checked exceptions as well
+   */
+  def exceptionalConsumer[T](f: ExceptionalJavaConsumer[T]) = new ExceptionalFunction[T, Unit] {
+    @throws(classOf[Throwable])
+    override def applyE(value: T): Unit = f(value)
+  }
+
+  /**
+   * Short form of `exceptionalConsumer`
+   */
+  def xc[T](f: ExceptionalJavaConsumer[T]) : ExceptionalFunction[T, Unit] = exceptionalConsumer(f)
 }
 
 abstract class ExceptionalFunction[-T1, +R] extends Function[T1, R] {
--- a/util-core/src/test/java/com/twitter/util/FunctionCompilationTest.java
+++ b/util-core/src/test/java/com/twitter/util/FunctionCompilationTest.java
@@ -1,6 +1,12 @@
 package com.twitter.util;
 
 import org.junit.Test;
+import scala.runtime.BoxedUnit;
+
+import static com.twitter.util.Function.c;
+import static com.twitter.util.Function.f;
+import static com.twitter.util.Function.xc;
+import static com.twitter.util.Function.xf;
 
 /**
  * Tests are not currently run for java, but for our purposes, if the test compiles at all, it's
@@ -9,7 +15,7 @@ import org.junit.Test;
 public class FunctionCompilationTest {
 
   /** Confirm that we can extend ExceptionalFunction with applyE(). */
-  @Test
+  @Test(expected = Exception.class)
   public void testDefineWithException() {
     ExceptionalFunction<Integer, String> fun = new ExceptionalFunction<Integer, String>() {
       @Override
@@ -17,16 +23,11 @@ public class FunctionCompilationTest {
         throw new Exception("Expected");
       }
     };
-    try {
-      fun.apply(1);
-      assert false : "Should have thrown";
-    } catch (Exception e) {
-      // pass: expected
-    }
+    fun.apply(1);
   }
 
   /** Confirm that we can extend ExceptionalFunction0 with applyE(). */
-  @Test
+  @Test(expected = Exception.class)
   public void testExceptionalFunction0() {
     ExceptionalFunction0<Integer> fun = new ExceptionalFunction0<Integer>() {
       @Override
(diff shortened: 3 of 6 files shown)


util-core/src/test/java/com/twitter/util/FunctionCompilationTest.java
```diff
@@ -1,6 +1,13 @@
 package com.twitter.util;
 
 import org.junit.Test;
+import scala.Unit;
+import scala.runtime.BoxedUnit;
+
+import static com.twitter.util.Function.func;
```
mosesn: sort imports

util-core/src/main/java/com/twitter/util/ExceptionalJavaConsumer.java
```diff
@@ -0,0 +1,10 @@
+package com.twitter.util;
+
+/**
+ * Used for Java 8 interop.
+ * This is a SAM version of c.t.u.ExceptionalFunction (returning Unit/void).
+ */
+// @FunctionalInterface
+public interface ExceptionalJavaConsumer<T> {
```
mosesn: is this even worth it given that we can't use FunctionalInterface?  Maybe it would make sense to publish a separate util-lambda target for now?

util-core/src/test/java/com/twitter/util/FunctionCompilationTest.java
```diff
@@ -34,11 +36,66 @@ public Integer applyE() throws Exception {
...
-      fun.apply();
-      assert false : "Should have thrown";
-    } catch (Exception e) {
-      // pass: expected
-    }
+    fun.apply();
+  }
+
+  @Test
+  public void testMakeFunctionFromLambda() {
```
mosesn: is it OK that these are just testing compilation?  we might stop running them and just compile them at some point in time.

mosesn:
This is very cool, thanks for the contribution!

Does this work even with the annotations commented out?  I'm a little worried that we'll generate java7 incompatible java byte code.  Other than that, I'm really excited about all of this.

util-core/src/main/java/com/twitter/util/ExceptionalJavaConsumer.java
eirslett: `@FunctionalInterface` is just a marker annotation, intended to inform the reader that the intention of the interface is to act as a SAM. With the Java 8 compiler, it also does some compile-time checks to ensure you can use the interface as a SAM (it should have only one non-default method).
For the lambda functionality itself, it has no consequences, since lambda annotations also work with non-annotated interfaces. It's only there to communicate intent.
I've commented it out since the annotation itself only exists in the Java 8 library (so that twitter-util is compatible with Java 7)

util-core/src/test/java/com/twitter/util/FunctionCompilationTest.java
eirslett: I think that's how they already work - they are compiled but not run. I commented out the Java 8 part since the code needs to be able to be compiled with JDK 7. (But if you un-outcomment it, then it should compile.) I was asking on [stackoverflow](http://stackoverflow.com/questions/27881595/how-to-set-up-build-to-accept-src-test-java8-only-when-java-8-is-used) if you can configure sbt to only compile java 8 files if running with jdk 8, so you can set up a Travis CI cross-build - until then, we cannot have actual lambda expressions here (unfortunately).

eirslett:
Yes, it works in a forwards-compatible manner. You can use the code from Java 7, but it will be more noisy.
If this code compiles on a jdk 7, you should be fairly sure it generates java-7-compatible bytecode! :-)

util-core/src/test/java/com/twitter/util/FunctionCompilationTest.java
eirslett: Ok, I fixed it!

mosesn:
LGTM!  I'll make sure we get some more eyes on this soon.

util-core/src/main/scala/com/twitter/util/Function.scala
```diff
@@ -47,6 +47,38 @@ object Function {
...
+   * Useful for e.g. future.onSuccess
+   */
+  def cons[T](f: JavaConsumer[T]) = new Function[T, Unit] {
+    override def apply(value: T): Unit = f(value)
+  }
+
+  /**
+   * like func, but deals with checked exceptions as well
+   */
+  def exfunc[T, R](f: ExceptionalJavaFunction[T, R]) = new ExceptionalFunction[T, R] {
```
mosesn: oh, I missed this before, but maybe camelcase these two for consistency?

vkostyukov:
All of this seem reasonable for me. Just one concern. As far as I know, you don't really need doing anything magical to make your interface a _functional_ interface in Java 8. @eirslett why `flatMap` for example, would not accept lambda function in Java 8 w/o any modifications? Is it because `c.t.u.Function` also bas methods `compose` and `andThen`?

eirslett:
Lambdas only work with interfaces. Since c.t.u.Function is an abstract class, you cannot use it as a lambda.

util-core/src/main/scala/com/twitter/util/Function.scala
eirslett: Do you mean like "exFunc" and "exCons"? We could certainly do that. It conforms to the java/scala naming conventions. My only concern is that the camel casing itself might (possibly) look a (tiny) bit noisier than the plain lowercased versions:

```java
Future.value(3)
  .flatMap(exfunc(i -> ...))
  .map(exfunc(i -> ...)
  .onSuccess(excons(...));
```

or

```java
Future.value(3)
  .flatMap(exFunc(i -> ...))
  .map(exFunc(i -> ...))
  .onSuccess(exCons(..));
```

The intention is to let the lambda contents shine and hide all the noise, so to speak. With that argument, they could be shortened further:

```java
Future.value(3)
  .flatMap(xf(i -> ...))
  .map(xf(i -> ...))
  .onSuccess(xc(..));
```

with "f", "c", "xf" and "xc", but then again, that would probably make the code less understandable if you don't know the conventions.

Do you still want me to camelCase to exFunc and exCons?

mosesn: how about we compromise and have two versions of the methods, one long / explicit, and one short / obscure?  that's a common pattern in scala if you want to have a symbolic DSL.  like maybe,

exceptionalFunction / xf
function / fn
exceptionalConstant / xc
constant / cn

stevegury: Java conventions are usually pretty verbose compared to Scala, so I would agree with Moses here.

stevegury:
Thanks Eirik for doing this!

util-core/src/main/scala/com/twitter/util/Function.scala
eirslett: Ok, I defined these methods:
exceptionalFunction / xf
function / f
exceptionalConsumer / xc
consumer / c

So, just "f" and "c".

mosesn:
LGTM, :ship:

stevegury:
LGTM, I will pull this internally.

eirslett:
@stevegury I'm not sure why it doesn't compile... looks like it succeeds on some builds, and sbt fails to start on others?

eirslett:
Closing/reopening to trigger a new Travs build.

mosesn:
@eirslett I'm trying to pull this in, but I just realized that it introduces a cycle between util-core-java and util-core-scala, which we need to avoid for historical reasons internally.  I'm going to split out the core java functions into util-java-function, ok?

mosesn:
The other option is that you can do it yourself and I'll merge in your change instead.

util-core/src/main/scala/com/twitter/util/Function.scala
mariusae: I'd much prefer to have just one, simple, name for these. e.g., I liked `func`. (`fun` is also good.)

mosesn: My argument for the short / DSL-y version is that for java users interoperating with scala, they'll probably use these methods all the time.  DSLs work well for those kinds of users.  On the other hand, for folks who only rarely need to use this lib, it's simpler for the name to explain what it's for.  I don't think alias-ing is such a sin.

mariusae: > On Feb 10, 2015, at 10:20 AM, Moses Nakamura <notifications@github.com> wrote:
> 
> In util-core/src/main/scala/com/twitter/util/Function.scala <https://github.com/twitter/util/pull/122#discussion_r24435731>:
> 
> > +  def func[T, R](f: JavaFunction[T, R]) = new Function[T, R] {
> > +    override def apply(value: T): R = f(value)
> > +  }
> > +
> > +  /**
> > +   * Creates a T => Unit from a JavaConsumer.
> > +   * Useful for e.g. future.onSuccess
> > +   */
> > +  def cons[T](f: JavaConsumer[T]) = new Function[T, Unit] {
> > +    override def apply(value: T): Unit = f(value)
> > +  }
> > +
> > +  /**
> > +   * like func, but deals with checked exceptions as well
> > +   */
> > +  def exfunc[T, R](f: ExceptionalJavaFunction[T, R]) = new ExceptionalFunction[T, R] {
> My argument for the short / DSL-y version is that for java users interoperating with scala, they'll probably use these methods all the time. DSLs work well for those kinds of users. On the other hand, for folks who only rarely need to use this lib, it's simpler for the name to explain what it's for. I don't think alias-ing is such a sin.
> 
I don’t agree. Why introduce this extra inconsistency when we don’t have to?

mosesn: For exactly the same reason why the dispatch library now has explicit operators as well as symbolic ones.  Obscure names are OK for folks who use a library all the time, but just make code harder to read if you only use a library every once in a while.  We do this with our Time and Duration APIs too.

With that said, I don't think this is worth fighting over, and I'd rather get it in, so I'm OK with func.  @mariusae @eirslett are you OK with my original request to camelcase?

mariusae: I'm not sure that we should take design hints from dispatch. I really don't think that it's a good idea to introduce aliases without really good justification. I personally liked the original ones. (`func`, `exfunc`, `cons`.)

mosesn: OK, let's keep the original ones.  I'll make the change on my side.

mosesn:
This finally made it into internal!  It should hit develop in ~ a week.  Thanks a lot for the PR :+1:

luciferous:
Closed by 016206c.

PR #124, opened 2015-01-22
Easy argument capture for Mockito mocks.

zacharyvoase:
## Problem
Mockito's ArgumentCaptor enables you to capture the arguments that were
passed to stubbed-out methods on mocks, to make assertions on them.
However, the API for ArgumentCaptor was built for Java and is kind of
clunky compared to what we expect in Scala.

An alternative to using ArgumentCaptor is to use Mockito argument
matchers in `when(myObj.method(...))`. However, this means every
assertion you want to make has to be encapsulated in a custom matcher
subclass, and you can't make assertions spanning more than one argument
in each stubbed method invocation. For example, you might want to check
that a key-value store wrapper object was called with a key of "foo" and
a timeout of 10 seconds, and Mockito argument matchers would not be able
to provide this.

Another alternative is to use matchers in a
`verify(myMock, times(N)).someMethod(...)` call, but this suffers from
the same problem of having to write a specific matcher subclass for any
non-standard assertions.

## Solution
The ArgumentCapture trait can be mixed into your test class and has two
methods, `capturingOne` and `capturingAll`, which capture the
invocations of your stubbed methods as a tuple or Seq of tuples. Here's
an example:

    when(myObj.lookup(any[String], any[Duration])).thenReturn(someValue)

    // The call to myObj.lookup will normally live somewhere inside the
    // subject under test.
    assert(myObj.lookup("foo", 10.seconds) === someValue)

    val (key, timeout) = capturingOne(verify(myObj).lookup _)
    assert(key === "foo")
    assert(timeout === 10.seconds)

An example which uses `capturingAll` and doesn't make any assertions
about the order of the stubbed method calls (`forExactly` is defined on
the `org.scalatest.Inspectors` trait):

    val requests = capturingAll(verify(myObj, times(2)).lookup _)

    forExactly(1, requests) { (key, timeout) =>
      assert(key === "foo")
      assert(timeout === 10.seconds)
    }

    forExactly(1, requests) { (key, timeout) =>
      assert(key === "bar")
      assert(timeout === 2.seconds)
    }

The only issue with this solution is that it currently only has
implementations up to 5-ary functions; vanilla Scala doesn't have
variadic type parameters so the captureOne/captureAll functions for each
arity have to be copied, pasted and extended by hand currently.


zacharyvoase:
BTW I ran all the tests with `./sbt test` and they passed.

vkostyukov:
@zacharyvoase could you please run `./sbt +test` to make sure we don't break Scala 2.10 build? Seems that there is one test failed: https://travis-ci.org/twitter/util/jobs/47861916#L1204, which is probably not related to this PR. Anyway, you should be able to check locally if everything works with `./sbt +test`.

zacharyvoase:
The test that failed doesn't look related to this PR, but I'm running `./sbt +test` (with the `+`) now.

zacharyvoase:
I get a failure in `com.twitter.jvm.ContentionTest` saying:

    The code passed to eventually never returned normally. Attempted 1 times over 3.0963380000000003 seconds. Last failure message: 1 did not equal 2. (ContentionTest.scala:48)

Probably unrelated, right?

vkostyukov:
Is this failure reproducible if you run the tests one more time. Although, Travis CI says it's green.

zacharyvoase:
Oh, they passed when I ran it a second time.

util-test/src/main/scala/com/twitter/util/testing/ArgumentCapture.scala
```diff
@@ -0,0 +1,115 @@
...
+import org.mockito.ArgumentCaptor
+import org.mockito.exceptions.Reporter
+import scala.collection.JavaConversions._
+import scala.reflect._
+
+trait ArgumentCapture {
+  /**
+   * Enables capturingOne to be implemented over capturingAll with the same behavior as ArgumentCaptor.getValue
+   */
+  private[this] def noArgWasCaptured: Nothing = {
```
mosesn: this doesn't act like a field, can you add parens?

util-test/src/main/scala/com/twitter/util/testing/ArgumentCapture.scala
```diff
@@ -0,0 +1,115 @@
...
+    new Reporter().noArgumentValueWasCaptured() // this always throws an exception
+    throw new RuntimeException("this should be unreachable, but allows the method to be of type Nothing")
+  }
+
+  private[this] def zip4[T, U, V, W](args: (JList[T], JList[U], JList[V], JList[W])): Seq[(T, U, V, W)] = {
```
mosesn: why are 4 and 5 special?

util-test/src/main/scala/com/twitter/util/testing/ArgumentCapture.scala
```diff
@@ -0,0 +1,115 @@
+package com.twitter.util.testing
+
+import java.util.{List => JList}
+import org.mockito.ArgumentCaptor
+import org.mockito.exceptions.Reporter
+import scala.collection.JavaConversions._
```
mosesn: in general, we prefer JavaConverters to JavaConversions.  If you want to avoid the extra allocation, could you invoke it explicitly?

util-test/src/main/scala/com/twitter/util/testing/ArgumentCapture.scala
mosesn: oh, I see, zipped works for 2 and 3 but not above.  what a hassle.

zacharyvoase: If we ever wanna extend this to 22-ary functions (for example) I'd just write a Python script which generated all of the capturing and zipping functions :-)

zacharyvoase: Or maybe macros would work? Not sure what constraints we have on stuff like that though.

util-test/src/main/scala/com/twitter/util/testing/ArgumentCapture.scala
```diff
@@ -0,0 +1,115 @@
...
+    capturingAll[T, U, V, W, X](f).lastOption.getOrElse(noArgWasCaptured)
+
+  /**
+   * Capture all the invocations from a verify(mock).method(arg) call.
+   *
+   * Example:
+   *   val requests = capturingAll(verify(myAPIEndpoint, times(4)).authenticate _)
+   *   requests.length must equal (4)
+   */
+  def capturingAll[T: ClassTag](f: T => _): Seq[T] = {
```
mosesn: could you include the script you wrote to generate these, in case we ever want to extend this or modify it ever?  ditto for capturingOne.

zacharyvoase: I wrote these by hand…haha

zacharyvoase: I could totally write a script though. Where would that live?

mosesn: for Future#join we did it like this: https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/util/Future.scala#L172

Although for 5 it's probably OK to omit it, not a huge deal.

util-test/src/main/scala/com/twitter/util/testing/ArgumentCapture.scala
mosesn: Yeah, I heard that pants is weird about macros?  Not sure whether that's apocryphal or not.  This is probably fine for now.

mosesn:
LGTM except for a couple style nits.  Thanks, I'm looking forward to using this :+1:

util-test/src/main/scala/com/twitter/util/testing/ArgumentCapture.scala
zacharyvoase: I wrote a script in Python to generate up to arities of 22, and everything works, but I'm not sure if it's small enough to put in the comments. I'll give it a go.

zacharyvoase:
I pushed the version supporting arities up to 21, and I have a really janky Python script that generates the Scala but it's not going to fit in the comments…

mosesn:
Maybe make a `gen` top level directory and stick your python script in there?

zacharyvoase:
In the top level of util? I'd be down

mosesn:
Yeah, that's what I was thinking.

zacharyvoase:
OK so I actually just built out a Make + Python based codegen system for the entire repo, it's all contained in codegen and hopefully it's useful although it means the scope of this PR just grew massively.

mosesn:
LGTM, although I'd definitely like to hear what other folks feel about this.

eirslett:
Would this be beneficial to other Scala projects as well? If then, maybe it could be even better to put it in a "mockito-scala" library?

mosesn:
@eirslett What are your concerns?  Is it mostly that util-test has unnecessary deps coming in from util-core and util-logging?

We could split it out into a util-mockito subproject, although I agree there isn't a particularly pressing reason for it to be in util (except perhaps that it fits with twitter/util's policy of neat knick-knacks).  I think the real benefit is that util already has its open source + publishing story down pat, so we won't have to go through the open source rigmarole either at twitter or with sonatype.

eirslett:
Yes, you're right about that. The easy solution might be just to suggest people use twitter-util if they want the enhanced Mockito features.

util-test/src/main/scala/com/twitter/util/testing/ArgumentCapture.scala
vkostyukov: Why is called `capturing` rather then `capture`? The same question is about `capturingOne`.

vkostyukov:
LGTM rather then tiny nitpicking about names.

util-test/src/main/scala/com/twitter/util/testing/ArgumentCapture.scala
zacharyvoase: I was going with the same vibe as Ruby's `returning`; it runs the block it's given and returns some output based on the block. In English I parse the code as 'capturing one [argument call], verify that obj.someMethod was called'

vkostyukov: Thanks for the explanation! Sounds reasonable to me.

vkostyukov:
LGTM :+1: after @zacharyvoase's explanation about names.

stevegury:
LGTM

mosesn:
I'm going to pull this internally.  I'll add you to the review, thanks @zacharyvoase :heart_eyes_cat:!

mosesn:
This was merged in!  Woo!

https://github.com/twitter/util/commit/cd2c09652f230f5fa63cfcfffc4a2e19ff5f25eb

PR #128, opened 2015-02-14
Fix race condition in ZkAsyncSemaphore

mikepii:
Before acquiring a permit, ZkAsyncSemaphore gets the list of existing permits and then checks each permit ZNode’s data. This raised an error when one of the existing permits is released between the time the list was gotten and the data was checked. This was an issue for us because we are using ZkAsyncSemaphore in Hadoop containers. When many containers are closing while others are opening, many permits get released rapidly.

This PR changes the shard permit consensus logic to count lost existing permits as having not been seen in the first place.

### DIFF
--- a/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
+++ b/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
@@ -262,7 +262,7 @@ class ZkAsyncSemaphore(zk: ZkClient, path: String, numPermits: Int, maxWaiters: 
     path.substring(permitNodePathPrefix.length).toInt
   }
 
-  private[this] def numPermitsOf(node: ZNode): Future[Int] = {
+  private[coordination] def numPermitsOf(node: ZNode): Future[Int] = {
     node.getData().transform {
       case Return(data: ZNode.Data) =>
         try {
--- a/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
+++ b/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
@@ -7,8 +7,8 @@ import org.apache.zookeeper.{CreateMode, KeeperException}
 import org.apache.zookeeper.KeeperException.NoNodeException
 
 import com.twitter.concurrent.Permit
-import com.twitter.util.{Throw, Return, Future, Promise}
-import com.twitter.zk.{StateEvent, ZNode, ZkClient}
+import com.twitter.util.{Future, Promise, Return, Throw}
+import com.twitter.zk.{StateEvent, ZkClient, ZNode}
 
 /**
  * ZkAsyncSemaphore is a distributed semaphore with asynchronous execution.
--- a/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
+++ b/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
@@ -266,11 +266,13 @@ class ZkAsyncSemaphore(zk: ZkClient, path: String, numPermits: Int, maxWaiters: 
     node.getData().transform {
       case Return(data: ZNode.Data) =>
         try {
-          Future(new String(data.bytes, Charset.forName("UTF8")).toInt)
+          Future.value(new String(data.bytes, Charset.forName("UTF8")).toInt)
         } catch {
-          case err: NumberFormatException => Future(-1)
+          case err: NumberFormatException => Future.value(-1)
         }
-      case Throw(t: NoNodeException) => Future(-1)
+      case Throw(t: NoNodeException) =>
+        // This permit was released (i.e. after we got the list of permits).
+        Future.value(-1)
       case Throw(t) => Future.exception(t)
     }
   }
--- a/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
+++ b/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
@@ -270,8 +270,8 @@ class ZkAsyncSemaphore(zk: ZkClient, path: String, numPermits: Int, maxWaiters: 
         } catch {
           case err: NumberFormatException => Future(-1)
         }
-      case Throw(t) =>
-        Future(-1)
+      case Throw(t: NoNodeException) => Future(-1)
+      case Throw(t) => Future.exception(t)
     }
   }
 
--- a/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
+++ b/util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
@@ -4,9 +4,10 @@ import java.nio.charset.Charset
 import java.util.concurrent.{ConcurrentLinkedQueue, RejectedExecutionException}
 
 import org.apache.zookeeper.{CreateMode, KeeperException}
+import org.apache.zookeeper.KeeperException.NoNodeException
 
 import com.twitter.concurrent.Permit
-import com.twitter.util.{Future, Promise}
+import com.twitter.util.{Throw, Return, Future, Promise}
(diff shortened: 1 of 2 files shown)


vkostyukov:
Looks good to me! Thanks @mikepii for fixing this. Let me get some more eyes on it.

util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
```diff
@@ -262,12 +263,15 @@ class ZkAsyncSemaphore(zk: ZkClient, path: String, numPermits: Int, maxWaiters:
...
-    node.getData() map { data =>
-      try {
-        new String(data.bytes, Charset.forName("UTF8")).toInt
-      } catch {
-        case err: NumberFormatException => -1
-      }
+    node.getData().transform {
+      case Return(data: ZNode.Data) =>
+        try {
+          Future(new String(data.bytes, Charset.forName("UTF8")).toInt)
```
mosesn: prefer Future.value, which doesn't require allocating an extra closure, especially if you're already in a try/catch block.

util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
```diff
@@ -262,12 +263,15 @@ class ZkAsyncSemaphore(zk: ZkClient, path: String, numPermits: Int, maxWaiters:
...
+        } catch {
+          case err: NumberFormatException => Future(-1)
```
mosesn: ditto

util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
```diff
@@ -262,12 +263,15 @@ class ZkAsyncSemaphore(zk: ZkClient, path: String, numPermits: Int, maxWaiters:
...
+        }
+      case Throw(t: NoNodeException) => Future(-1)
```
mosesn: ditto

mosesn:
Could you follow [CONTRIBUTED.md][0]?

[0]: https://github.com/twitter/util/blob/master/CONTRIBUTING.md#style

util-zk/src/test/scala/com/twitter/zk/coordination/ZkAsyncSemaphoreTest.scala
```diff
@@ -143,6 +144,28 @@ class ZkAsyncSemaphoreTest extends WordSpec with MockitoSugar with AsyncAssertio
...
+          withClient { zk =>
+            val sem = new ZkAsyncSemaphore(zk, "/aoeu/aoeu", 2)
+            val znode = ZNode(zk, "/testing/twitter/uuu")
+            znode.create("7".getBytes)
+            val permits: Future[Int] = sem invokePrivate numPermitsOfMethod(znode)
+            assert(permits.get() == 7)
+          }
+        }
+
+        "not error on NoNode" in {
```
mosesn: this doesn't really test what we care about, right?  which is that acquire shouldn't fail just because you have a node that doesn't exist.

util-zk/src/test/scala/com/twitter/zk/coordination/ZkAsyncSemaphoreTest.scala
```diff
@@ -11,15 +11,16 @@ import org.scalatest.concurrent.AsyncAssertions
 import org.scalatest.concurrent.Eventually._
 import org.scalatest.junit.JUnitRunner
 import org.scalatest.mock.MockitoSugar
+import org.scalatest.PrivateMethodTester
```
mosesn: instead of using this, we can just widen the scope of numPermitsOf to `private[coordination]`, I think that's simpler.  Do you have a strong preference for using PrivateMethodTester?

util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
mosesn: could you add a comment which explains why this might happen?

mosesn:
I'm having a lot of trouble understanding how `ZkAsyncSemaphore` works at all.  It seems totally broken to me.  I'll try to take a closer look this weekend.  @mikepii if you could document the strategy behind `ZkAsyncSemaphore`, and add that to the review, it would help this go faster and also be a very useful general improvement.

mosesn:
OK, I figured it out.  @mikepii is there any way that you can also remove the failed nodes from the list of permits before we grab sequence numbers?  LGTM otherwise.

mikepii:
@mosesn, thanks for the detailed feedback. I've started incorporating some of it:
- got rid of `PrivateMethodTester` and expanded the private scope of `numPermitsOf`
- replaced `Future.apply` with `Future.value`
- added the comment to explain what we're doing with the `NoNodeException`
- some style things (I haven't had a chance to read those style guides thoroughly yet but I found a couple of the issues so far)
- I will re-format the commit message at the end

I also found an issue in the regression test I added and corrected that.

In response to your concern about the other test -- it does fail if `NoNode` is raised when reading the permit, which is where the issue was (i.e. it fails if you checkout master ZkAsyncSemaphore.scala). Would you prefer that we test this higher in the stack at either `getConsensusNumPermits` or `acquire`? I changed the names of the nodes to make it more clear what it's testing.

I need some more time to consider what you said about removing failed nodes from the list of permits before grabbing sequence numbers.

I'm making new changes on [fix/shard-race](https://github.com/RUNDSP/util/commits/fix/shard-race). My plan is to reset this PR branch and then squash those commits to replace the single PR commit. Is that alright? Will we lose your line comments above on the PR because of the reset?

mosesn:
Yeah, I'd prefer that it's at acquire.  My concern is that someone will refactor the class, and because we're testing this private method, if the private method disappears, or starts being used differently, the test will no longer protect us from this kind of regression.  I think testing at the acquire level is not that hard, and demonstrates what we want better.  Does that make sense?

Re resetting the PR branch: I don't think so, but even if we do it's fine.  In general, I think github is good about accommodating many different kinds of git workflows.  With that said, we squash your commits on our side anyway, so if you want to just merge your changes in, that's totally fine.  It would be nice to have a commit message which follows Problem / Solution / Result (makes our lives a little easier) but if you just supply the text that's fine too.

mosesn:
@mikepii howdy, do you need any further help from us?

mikepii:
@mosesn, thanks, but I can handle it. Sorry this is taking so long, I haven't been able to find any free time to work on it. Hopefully this week.

util-zk/src/main/scala/com/twitter/zk/coordination/ZkAsyncSemaphore.scala
```diff
@@ -262,12 +263,15 @@ class ZkAsyncSemaphore(zk: ZkClient, path: String, numPermits: Int, maxWaiters:
...
 
   private[this] def numPermitsOf(node: ZNode): Future[Int] = {
-    node.getData() map { data =>
-      try {
-        new String(data.bytes, Charset.forName("UTF8")).toInt
-      } catch {
-        case err: NumberFormatException => -1
-      }
+    node.getData().transform {
+      case Return(data: ZNode.Data) =>
```
luciferous: Is data ever not ZNode.Data?

luciferous:
Checking in again – anything we can do to help out?

mikepii:
Sorry this branch is getting stale, just rebased it. Actually, it's not clear to me how to test this at the `acquire` level, since the situation we are trying to reproduce happens when a node is deleted *while* `acquire` is running.

mosesn:
Sorry this has taken so long for me to get back to you.  I'm going to need some time to take a look at this again because it's always good to be careful around concurrent code (and doubly so for anything ZK-y), and I'm spending most of my time on a different project at work.  If you need a faster turnaround, I can try to get someone else who has more bandwidth to take a look.

mosesn:
Agh, I'm the worst, but this looks good to me.  I deeply apologize for how long it took me to review this.

mosesn:
This was merged here: https://github.com/twitter/util/commit/f78382ef261e851261f3f541ca94743f7f7f0435.  Thanks @mikepii!

PR #131, opened 2015-03-24
Replace reference to scala.ScalaObject with scala.AnyVal

kpfell:
Currently, `util-eval` seems not to be usable in Scala 2.11, since `ScalaObject` was deprecated in 2.10.

Replacing the reference to `ScalaObject` with `AnyVal` should work on both 2.10 and 2.11, though?

### DIFF
--- a/util-eval/src/main/scala/com/twitter/util/Eval.scala
+++ b/util-eval/src/main/scala/com/twitter/util/Eval.scala
@@ -80,7 +80,7 @@ class Eval(target: Option[File]) {
   }
 
   private lazy val libPath = try {
-    classPathOfClass("scala.ScalaObject")
+    classPathOfClass("scala.AnyVal")
   } catch {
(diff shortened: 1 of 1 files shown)


travisbrown:
Thanks, @kpfell. This is precisely the fix we're using in the [`develop` branch](https://github.com/twitter/util/commit/84b9fa50916b496f71d2fd9b0cbb34f191b8a51f), and it'll be available when we publish a 6.24.0 release.

When we began publishing the `develop` branch on a weekly basis we decided to keep `master` the default on GitHub since it tracks releases, but given that releases can be _very_ far apart I'm starting to reconsider this decision. In any case thanks for the fix, and sorry it wasn't clearer that it had already been implemented.

kpfell:
@travisbrown Ah, alright!

One day, I will be able to use this library. :sparkles: _One day...!_ :sparkles:

PR #135, opened 2015-06-26
Add facility to round up to a duration in TimeLike

benjumanji:
Problem

No rounding operation analgous to floor for rounding up to the closest
multiple of a duration. While less common than floor, ceiling is still
useful.

Solution

Add a ceiling function to TimeLike.

### DIFF
--- a/util-core/src/main/scala/com/twitter/util/Duration.scala
+++ b/util-core/src/main/scala/com/twitter/util/Duration.scala
@@ -480,5 +480,7 @@ sealed class Duration private[util] (protected val nanos: Long) extends {
    */
   def rem(x: Duration): Duration = this % x
 
-  override def floor(x: Duration): Duration = super.floor(x) // for Java-compatibility
+  // for Java-compatibility
+  override def floor(x: Duration): Duration = super.floor(x)
+  override def ceil(x: Duration): Duration = super.ceil(x)
 }
--- a/util-core/src/main/scala/com/twitter/util/Duration.scala
+++ b/util-core/src/main/scala/com/twitter/util/Duration.scala
@@ -481,6 +481,6 @@ sealed class Duration private[util] (protected val nanos: Long) extends {
   def rem(x: Duration): Duration = this % x
 
   // for Java-compatibility
-  override def floor(x: Duration): Duration = super.floor(x)
-  override def ceil(x: Duration): Duration = super.ceil(x)
+  override def floor(increment: Duration): Duration = super.floor(increment)
+  override def ceil(increment: Duration): Duration = super.ceil(increment)
 }
--- a/util-core/src/main/scala/com/twitter/util/Time.scala
+++ b/util-core/src/main/scala/com/twitter/util/Time.scala
@@ -195,19 +195,29 @@ trait TimeLike[This <: TimeLike[This]] extends Ordered[This] { self: This =>
   /** The difference between the two `TimeLike`s */
   def diff(that: This): Duration
 
+  private[this] def round(f: (Long, Long) => Long, x: Duration): This = (this, x) match {
+    case (Nanoseconds(0), Duration.Nanoseconds(0)) => Undefined
+    case (Nanoseconds(num), Duration.Nanoseconds(0)) => if (num < 0) Bottom else Top
+    case (Nanoseconds(num), Duration.Nanoseconds(denom)) => fromNanoseconds(f(num, denom))
+    case (self, Duration.Nanoseconds(_)) => self
+    case (_, _) => Undefined
+  }
+
+  /**
+   * Rounds up to the nearest multiple of the given duration.  For example:
+   * 127.seconds.ceiling(1.minute) => 3.minutes.  Taking the ceiling of a
+   * Time object with duration greater than 1.hour can have unexpected
+   * results because of timezones.
+   */
+  def ceiling(x: Duration): This = round((num, denom) => ((num + denom - 1)/denom) * denom, x)
+
   /**
    * Rounds down to the nearest multiple of the given duration.  For example:
    * 127.seconds.floor(1.minute) => 2.minutes.  Taking the floor of a
    * Time object with duration greater than 1.hour can have unexpected
    * results because of timezones.
    */
-  def floor(x: Duration): This = (this, x) match {
-    case (Nanoseconds(0), Duration.Nanoseconds(0)) => Undefined
-    case (Nanoseconds(num), Duration.Nanoseconds(0)) => if (num < 0) Bottom else Top
-    case (Nanoseconds(num), Duration.Nanoseconds(denom)) => fromNanoseconds((num/denom) * denom)
-    case (self, Duration.Nanoseconds(_)) => self
-    case (_, _) => Undefined
-  }
+  def floor(x: Duration): This = round((num, denom) => (num/denom) * denom, x)
 
   def max(that: This): This =
     if ((this compare that) < 0) that else this
--- a/util-core/src/main/scala/com/twitter/util/Time.scala
+++ b/util-core/src/main/scala/com/twitter/util/Time.scala
@@ -637,6 +637,8 @@ sealed class Time private[util] (protected val nanos: Long) extends {
    *  Finds a diff between this and ''that'' time.
    */
   def minus(that: Time): Duration = this - that
-
-  override def floor(x: Duration): Time = super.floor(x) // for Java-compatibility
+  
+  // for Java-compatibility
+  override def floor(x: Duration): Time = super.floor(x)
+  override def ceil(x: Duration): Time = super.ceil(x) 
 }
--- a/util-core/src/main/scala/com/twitter/util/Time.scala
+++ b/util-core/src/main/scala/com/twitter/util/Time.scala
@@ -20,6 +20,8 @@ import java.io.Serializable
 import java.util.concurrent.TimeUnit
 import java.util.{Date, Locale, TimeZone}
 
+import com.twitter.conversions.time.intToTimeableNumber
+
 /**
  * @define now
  *
@@ -201,7 +203,9 @@ trait TimeLike[This <: TimeLike[This]] extends Ordered[This] { self: This =>
    * Time object with duration greater than 1.hour can have unexpected
    * results because of timezones.
    */
-  def ceil(increment: Duration): This = floor(increment) + increment
+  // the impl might seem weird but consider the case where this is an exact multiple of the 
+  // increment.
+  def ceil(increment: Duration): This = (this - 1.nanoseconds).floor(increment) + increment
 
   /**
    * Rounds down to the nearest multiple of the given duration.  For example:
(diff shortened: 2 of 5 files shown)


vkostyukov:
This is cool @benjumanji!

Do you mind adding a Java compilation test for a new function to both `DurationCompilationTest`and `TimeCompilationTest` (see `util-core/src/test/java`). It looks to me that it won't accessible from Java, but you could easily workaround it like shown [here](https://github.com/twitter/util/blob/master/util-core/src/main/scala/com/twitter/util/Time.scala#L629).

travisbrown:
@vkostyukov We should probably add a note about Java compilation tests to our contributing guides.

util-core/src/main/scala/com/twitter/util/Time.scala
```diff
@@ -195,19 +195,29 @@ trait TimeLike[This <: TimeLike[This]] extends Ordered[This] { self: This =>
...
+    case (_, _) => Undefined
+  }
+
+  /**
+   * Rounds up to the nearest multiple of the given duration.  For example:
+   * 127.seconds.ceiling(1.minute) => 3.minutes.  Taking the ceiling of a
+   * Time object with duration greater than 1.hour can have unexpected
+   * results because of timezones.
+   */
+  def ceiling(x: Duration): This = round((num, denom) => ((num + denom - 1)/denom) * denom, x)
```
mosesn: do you mind if we call this `ceil` for consistency with scala.util.math?

mosesn: instead of `x` is there a more descriptive word?

util-core/src/main/scala/com/twitter/util/Time.scala
```diff
@@ -195,19 +195,29 @@ trait TimeLike[This <: TimeLike[This]] extends Ordered[This] { self: This =>
...
    * results because of timezones.
    */
-  def floor(x: Duration): This = (this, x) match {
-    case (Nanoseconds(0), Duration.Nanoseconds(0)) => Undefined
-    case (Nanoseconds(num), Duration.Nanoseconds(0)) => if (num < 0) Bottom else Top
-    case (Nanoseconds(num), Duration.Nanoseconds(denom)) => fromNanoseconds((num/denom) * denom)
-    case (self, Duration.Nanoseconds(_)) => self
-    case (_, _) => Undefined
-  }
+  def floor(x: Duration): This = round((num, denom) => (num/denom) * denom, x)
```
mosesn: while we're in here, can we rename x to something better?

mosesn:
Can you val up the arguments to `ceiling` and `floor`?  Neither of these operations should entail an allocation for the closure.

util-core/src/main/scala/com/twitter/util/Time.scala
```diff
@@ -195,19 +195,29 @@ trait TimeLike[This <: TimeLike[This]] extends Ordered[This] { self: This =>
   /** The difference between the two `TimeLike`s */
   def diff(that: This): Duration
 
+  private[this] def round(f: (Long, Long) => Long, x: Duration): This = (this, x) match {
```
mosesn: the names of these arguments should be more descriptive

mosesn:
I'm not sure that I love this approach, where we pass a slightly complicated function to `round`.  With the closure, we're making something that used to be very simple a little complicated, and it doesn't seem easier to read.  Instead of adding the `round` method, could we instead define `ceiling` as

```scala
def ceil(increment: Duration): This = floor(increment) + increment
```

util-core/src/test/scala/com/twitter/util/TimeTest.scala
```diff
@@ -307,19 +302,36 @@ trait TimeLikeSpec[T <: TimeLike[T]] extends WordSpec with GeneratorDrivenProper
...
-    "floor itself" in {
+    "round to itself" in {
       for (s <- Seq(Long.MinValue, -1, 1, Long.MaxValue); t = fromNanoseconds(s))
         assert(t.floor(Duration.fromNanoseconds(t.inNanoseconds)) === t)
     }
   }
 
+  "floor" should {
+    "round down" in {
+      assert(fromSeconds(60).floor(1.minute) === fromSeconds(60))
```
mosesn: no need for === anymore, scalatest 2.x supplied an assert macro which unpacks failures very neatly even when we use regular old ==.

util-core/src/test/scala/com/twitter/util/TimeTest.scala
```diff
@@ -307,19 +302,36 @@ trait TimeLikeSpec[T <: TimeLike[T]] extends WordSpec with GeneratorDrivenProper
...
 
-    "floor itself" in {
+    "round to itself" in {
       for (s <- Seq(Long.MinValue, -1, 1, Long.MaxValue); t = fromNanoseconds(s))
         assert(t.floor(Duration.fromNanoseconds(t.inNanoseconds)) === t)
     }
   }
 
+  "floor" should {
+    "round down" in {
```
mosesn: feel free to ignore:
I think this is a tiny bit more readable if you use 60.seconds helpers rather than fromSeconds, but it's not a big deal (and fromSeconds will actually make it compile faster, so there's definitely an argument the other way too).

util-core/src/test/scala/com/twitter/util/TimeTest.scala
```diff
@@ -288,16 +288,11 @@ trait TimeLikeSpec[T <: TimeLike[T]] extends WordSpec with GeneratorDrivenProper
...
-      assert(fromSeconds(100).floor(1.minute) === fromSeconds(60))
-      assert(fromSeconds(119).floor(1.minute) === fromSeconds(60))
-      assert(fromSeconds(120).floor(1.minute) === fromSeconds(120))
-    }
-
+  "rounding" should {
+    
     "maintain top and bottom" in {
       assert(Top.floor(1.hour) === Top)
+      assert(Bottom.floor(1.second) === Bottom)
```
mosesn: feel free to ignore:
for symmetry, keep this 1.hour too?  no reason for people who are next looking at this test to wonder, "why is one 1.hour and the other 1.second?" when, as far as I know, there isn't a reason.

mosesn:
Thanks for the contribution, @benjumanji!

benjumanji:
Wow! Such a lot of feedback. I will get to this over the weekend probably.

util-core/src/main/scala/com/twitter/util/Time.scala
benjumanji: Done (I copied increment, not sure it's the best, but I didn't have anything better).

util-core/src/main/scala/com/twitter/util/Time.scala
benjumanji: Done.

benjumanji:
There is a slight wrinkle in the `def ceil(increment: Duration): This = floor(increment) + increment` implementation which can be handled by decrementing the value being rounded by one nano. This preserves the behaviour that rounding on an aligned boundary is a no-op. I'm not sure if I love it, but it does avoid and higher order function jiggery pokery.

util-core/src/main/scala/com/twitter/util/Duration.scala
```diff
@@ -480,5 +480,7 @@ sealed class Duration private[util] (protected val nanos: Long) extends {
    */
   def rem(x: Duration): Duration = this % x
 
-  override def floor(x: Duration): Duration = super.floor(x) // for Java-compatibility
+  // for Java-compatibility
+  override def floor(x: Duration): Duration = super.floor(x)
```
mosesn: can we call these increment too?

util-core/src/main/scala/com/twitter/util/Time.scala
```diff
@@ -20,6 +20,8 @@ import java.io.Serializable
 import java.util.concurrent.TimeUnit
 import java.util.{Date, Locale, TimeZone}
 
+import com.twitter.conversions.time.intToTimeableNumber
```
mosesn: please sort imports

util-core/src/main/scala/com/twitter/util/Time.scala
```diff
@@ -196,12 +198,22 @@ trait TimeLike[This <: TimeLike[This]] extends Ordered[This] { self: This =>
...
 
   /**
+   * Rounds up to the nearest multiple of the given duration.  For example:
+   * 127.seconds.ceiling(1.minute) => 3.minutes.  Taking the ceiling of a
+   * Time object with duration greater than 1.hour can have unexpected
+   * results because of timezones.
+   */
+  // the impl might seem weird but consider the case where this is an exact multiple of the 
+  // increment.
+  def ceil(increment: Duration): This = (this - 1.nanoseconds).floor(increment) + increment
```
mosesn: sorry to be anal about this, but slow libs are often slow because of death by a thousand cuts.  `1.nanoseconds` adds an allocation, can we val this up?

util-core/src/main/scala/com/twitter/util/Time.scala
```diff
@@ -625,6 +637,8 @@ sealed class Time private[util] (protected val nanos: Long) extends {
    *  Finds a diff between this and ''that'' time.
    */
   def minus(that: Time): Duration = this - that
-
-  override def floor(x: Duration): Time = super.floor(x) // for Java-compatibility
+  
+  // for Java-compatibility
+  override def floor(x: Duration): Time = super.floor(x)
```
mosesn: can we make these arguments increment?

util-core/src/main/scala/com/twitter/util/Time.scala
mosesn: feel free to ignore:
I think we can also avoid an allocation by checking if this.floor(increment) == this, and only adding increment if they're different.  If you think it decreases readability though, the allocation might be worth it.

benjumanji: Agree about the spurious allocation. I guess taking the diff will also generate a fresh object before applying the floor (jit malarky aside). I think the if check probably ends up being more readable, because all the comment is talking about is a trick to elide it.

mosesn:
LGTM

util-core/src/test/java/com/twitter/util/DurationCompilationTest.java
```diff
@@ -127,4 +127,12 @@ public void testFloor() {
 
     Assert.assertEquals(2, b.inSeconds());
   }
+
+  @Test
+  public void testCeil() {
```
vkostyukov: Perhaps adding the same test for `TimeCompilationTest`?

vkostyukov:
Thanks @benjumanji! This looks very good.

benjumanji:
Thanks for the feedback guys. Always a pleasure.

vkostyukov:
I'm pulling this for the internal review. Will post the update here once it gets merged internally.

vkostyukov:
This has been merged internally. It will show up on the `develop` branch early next week. Thanks @benjumanji!

Issue #138, opened 2015-09-08
AsyncStream.scanLeft is not eagerly satisfied with initial value.


olix0r:
attn: @luciferous 

When using `AsyncStream.scanLeft`, I expect the initial value, `z`, to be readable immediately.  It appears, however, that this value is not readable until the next value in the stream has been satisfied.

The docs lead me to believe that this is a bug:
```scala

  /**
   * Similar to foldLeft, but produces a stream from the result of each
   * successive fold:
   *
   * {{{
   * AsyncStream(1, 2, ...).scanLeft(z)(f) == z +:: f(z, 1) +:: f(f(z, 1), 2) +:: ...
   * }}}
   *
   * Note that for an `AsyncStream as`:
   *
   * {{{
   * as.scanLeft(z)(f).last == as.foldLeft(z)(f)
   * }}}
   */
```

When I construct a stream like:
```scala
AsyncStream.fromFuture(f).scanLeft(init)(_ + _)
```
I expect this stream to provide `init` immediately and then the following value once `f` has been satisfied.  I have instrumented a test demonstrate this behavior:
```scala
  test("AsyncStream.scanLeft is satisfied immediately with initial value") {
    val first = new Promise[Unit]
    val stream = AsyncStream.fromFuture(first map (_ => 1)).scanLeft(0)(_ + _)

    val uncons = stream.uncons
    //first.setDone()
    assert(uncons.isDefined)
    Await.result(uncons) match {
      case Some((hd, rest)) =>
        assert(hd == 0)
      case None => fail("no uncons")
    }
  }
```
when `first.setDone()` is uncommented, this test succeeds, otherwise `uncons.isDefined` fails.

PR #139, opened 2015-09-08
Let AsyncStream.scanLeft return values eagerly.

olix0r:
Problem

AsyncStream.scanLeft only provides a value once the next value in the stream
has been satisfied.  Notably, the initial value is not consumable until the
underlying stream has been satisfied.

Solution

Provide values eagerly.

Fixes twitter/util#138

### DIFF
--- a/util-core/src/test/scala/com/twitter/concurrent/exp/AsyncStreamTest.scala
+++ b/util-core/src/test/scala/com/twitter/concurrent/exp/AsyncStreamTest.scala
@@ -192,6 +192,13 @@ class AsyncStreamTest extends FunSuite with GeneratorDrivenPropertyChecks {
     }
   }
 
+  test("scanLeft is eager") {
+    val never = AsyncStream.fromFuture(Future.never)
+    val hd = never.scanLeft("hi")((_,_) => ???).head
+    assert(hd.isDefined)
+    assert(Await.result(hd) == Some("hi"))
+  }
+
   test("foldLeft") {
     forAll { (a: List[Int]) =>
       def f(s: String, n: Int) = (s.toLong + n).toString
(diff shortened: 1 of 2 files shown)


util-core/src/test/scala/com/twitter/concurrent/exp/AsyncStreamTest.scala
```diff
@@ -192,6 +192,37 @@ class AsyncStreamTest extends FunSuite with GeneratorDrivenPropertyChecks {
     }
   }
 
+  test("scanLeft is eager") {
+    val first = new Promise[Unit]
+
+    var stream = AsyncStream.fromFuture(first map (_ => 2)).scanLeft(1)(_ + _)
```
mosesn: ffti: the style is now to avoid infix except for symbolic operators, so this should be `first.map(_ => 2)`

util-core/src/test/scala/com/twitter/concurrent/exp/AsyncStreamTest.scala
```diff
@@ -192,6 +192,37 @@ class AsyncStreamTest extends FunSuite with GeneratorDrivenPropertyChecks {
...
+    var uncons = stream.uncons
```
mosesn: ffti: this might be simpler to reason about if we use different vals instead of a var.

mosesn:
LGTM.  It might be nice in the future to build a type of equality for AsyncStream to simplify this kind of test, where you can set which bits have been eagerly evaluated or not, and which asynchronous bits have been satisfied or not.

luciferous:
This seems reasonable to me. Let's say we have `never`, an AsyncStream which never produces (i.e. `fromFuture(Future.never)`), we want `never.scanLeft(a)(f) = a +:: never`.

What do you think of using this test instead?:

```
val as = fromFuture(Future.never).scanLeft(a)((_, _) => ???)
Await.result(as.head) == a
```

Can we update the scaladoc with the equality stated above. Feel free to reword.

olix0r:
I updated the branch with the requested changes. (I'm pre-squashing the branch before I push, so the full commit log isn't making into the review).

mosesn:
LGTM

luciferous:
LGTM

taylorleese:
@mosesn @luciferous Close this? Looks like the change is in already. https://github.com/twitter/util/commit/b169af7442188c8d6899eb294284bfbaea105564

luciferous:
We closed #138. Looks like this was a duplicate.

PR #143, opened 2015-10-09
Future.sequence

Chandler:
Add a method that allows you to execute an list of futures sequentially

Problem
Most methods that operate on lists of futures do so in parallel, because it's generally inefficient to sequence futures that don't depend on each other.

However there are some use cases for sequencing futures, such as executing side effects that must take place in order. 

Solution

Add a method `sequence` to Future that takes a list of functions that produce futures, and flatMaps them together. We take a list of functions because if you created a list of futures directly, they would already be executing in parallel.

Result

Now you can execute Futures in sequence.

### DIFF
--- a/util-core/src/main/scala/com/twitter/util/Future.scala
+++ b/util-core/src/main/scala/com/twitter/util/Future.scala
@@ -354,7 +354,7 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
    *     )
    *   Future.sequence(orderedOperations)
    * 
-   * @param fs a sequence of Function0's that returns Futures to be executed in sequence
+   * @param fs a sequence of Function0s that returns Futures to be executed in sequence
    * @return a `Future[Seq[A]]` containing the results of each future in fs
    */
   def sequence[A](fs: Seq[() => Future[A]]): Future[Seq[A]] = 
--- a/util-core/src/main/scala/com/twitter/util/Future.scala
+++ b/util-core/src/main/scala/com/twitter/util/Future.scala
@@ -358,7 +358,7 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
    * @return a `Future[Seq[A]]` containing the results of each future in fs
    */
   def sequence[A](fs: Seq[() => Future[A]]): Future[Seq[A]] = 
-    fs.foldLeft(Future(List.empty[A])) { (resultsFuture, nextFunction) =>
+    fs.foldLeft(Future(Vector.empty[A])) { (resultsFuture, nextFunction) =>
       for {
         results    <- resultsFuture
         nextResult <- nextFunction()
--- a/util-core/src/main/scala/com/twitter/util/Future.scala
+++ b/util-core/src/main/scala/com/twitter/util/Future.scala
@@ -339,20 +339,26 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
   def join[A](fs: JList[Future[A]]): Future[Unit] = Futures.join(fs)
 
   /**
-   * Take a sequence and sequentially apply a function f to each item.
-   * Then return all future results as a single Future[Seq[_]].
-   * 
-   * If during execution any f() throws an exception that exception will be returned and 
-   * the remaining futures will not be processed.
-   * 
+   * Take a sequence and sequentially apply a function `f` to each item.
+   * Then return all future results `as` a single Future[Seq[_]].
+   *
+   * If during execution any `f` is satsified `as` a failure ([[Future.exception]])
+   * then that failed Future will be returned and the remaining elements of `as`
+   * will not be processed.
+   *
    * usage:
-   *   Future.traverseSequentially(Seq(1,2,3))(deleteItem)
-   * 
-   * @param as a sequence of A that will have f applied to each item sequentially
-   * @return a `Future[Seq[B]]` containing the results of f being applied to every item in as
+   *  {{{
+   *    // will return a Future of `Seq(2, 3, 4)`
+   *    Future.traverseSequentially(Seq(1, 2, 3)) { i =>
+   *      Future(i + 1)
+   *    }
+   *  }}}
+   *
+   * @param `as` a sequence of `A` that will have `f` applied to each item sequentially
+   * @return a `Future[Seq[B]]` containing the results of `f` being applied to every item in `as`
    */
   def traverseSequentially[A,B](as: Seq[A])(f: A => Future[B]): Future[Seq[B]] =
-    as.foldLeft(Future(Vector.empty[B])) { (resultsFuture, nextItem) =>
+    as.foldLeft(Future.value(Vector.empty[B])) { (resultsFuture, nextItem) =>
       for {
         results    <- resultsFuture
         nextResult <- f(nextItem)
--- a/util-core/src/main/scala/com/twitter/util/Future.scala
+++ b/util-core/src/main/scala/com/twitter/util/Future.scala
@@ -339,43 +339,25 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
   def join[A](fs: JList[Future[A]]): Future[Unit] = Futures.join(fs)
 
   /**
-   * Take a sequence and sequentally apply a function f to each item.
+   * Take a sequence and sequentially apply a function f to each item.
    * Then return all future results as a single Future[Seq[_]].
    * 
    * If during execution any f() throws an exception that exception will be returned and 
    * the remaining futures will not be processed.
    * 
+   * usage:
+   *   Future.traverseSequentially(Seq(1,2,3))(deleteItem)
+   * 
    * @param as a sequence of A that will have f applied to each item sequentially
    * @return a `Future[Seq[B]]` containing the results of f being applied to every item in as
    */
-  def traverseSequentally[A,B](as: Seq[A])(f: A => Future[B]): Future[Seq[B]] =
+  def traverseSequentially[A,B](as: Seq[A])(f: A => Future[B]): Future[Seq[B]] =
     as.foldLeft(Future(Vector.empty[B])) { (resultsFuture, nextItem) =>
       for {
         results    <- resultsFuture
         nextResult <- f(nextItem)
       } yield (results :+ nextResult)
     }
-
-  /**
-   * Take a sequence of `() => Future[_]` and execute each
-   * future sequentially, then return all future results as a single Future[Seq[_]].
-   * 
-   * If during execution any of the futures throws an exception that exception will be returned and 
-   * the remaining futures will not be processed.
-   * 
-   * A use case for sequential processing is ordered operations: 
-   *   val orderedOperations =
-   *     Seq(
-   *       () => deleteItem(id),
-   *       () => undeleteItem(id),
-   *       () => deleteItem(id)
-   *     )
-   *   Future.sequenceEffects(orderedOperations)
-   * 
-   * @param fs a sequence of Function0s that returns Futures to be executed in sequence
-   * @return a `Future[Seq[A]]` containing the results of each future in fs
-   */
-  def sequenceEffects[A](fs: Seq[() => Future[A]]): Future[Seq[A]] = traverseSequentally(fs)(f => f())
 
   /**
    * Collect the results from the given futures into a new future of
--- a/util-core/src/main/scala/com/twitter/util/Future.scala
+++ b/util-core/src/main/scala/com/twitter/util/Future.scala
@@ -340,7 +340,7 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
 
   /**
    * Take a sequence and sequentally apply a function f to each item.
-   * future sequentially, then return all future results as a single Future[Seq[_]].
+   * Then return all future results as a single Future[Seq[_]].
    * 
    * If during execution any f() throws an exception that exception will be returned and 
    * the remaining futures will not be processed.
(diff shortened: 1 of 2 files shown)


luciferous:
This might be useful. I want to hear what others think about it. Note: we can implement this quite simply with AsyncStream.

```
def sequence[A](fs: Seq[() => Future[A]]): AsyncStream[A] =
  fs match {
    case Nil => AsyncStream.empty
    case head +: tail => AsycnStream.fromFuture(head()) ++ sequence(tail)
  }
```

If you want a `Future[Seq[A]]`, `sequence(fs).toSeq()` works. It's neater (imo) and we don't have to rely on the Seq's implementation of append (`:+`).

mosesn:
What's an example of side-effecting code you want to sequence?

stuhood:
Also, thinking about this more I wonder whether it would be more powerful as a `scanLeft`-alike... where each function receives the output of the previous.

util-core/src/main/scala/com/twitter/util/Future.scala
```diff
@@ -339,6 +339,33 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
...
+   *       () => deleteItem(id),
+   *       () => undeleteItem(id),
+   *       () => deleteItem(id)
+   *     )
+   *   Future.sequence(orderedOperations)
+   * 
+   * @param fs a sequence of Function0s that returns Futures to be executed in sequence
+   * @return a `Future[Seq[A]]` containing the results of each future in fs
+   */
+  def sequence[A](fs: Seq[() => Future[A]]): Future[Seq[A]] =
```
taylorleese: I'm not a fan of calling this ```sequence``` because ```sequence``` is typically understood to transform an ```M[Future[A]]``` to a ```Future[M[A]]```.

dschobel:
People have asked for some variant of this on internal finaglers a few times in recent memory so +1 to introducing it.

I think we can generalize to end up with what's known as the traverse function in other languages:

```scala
def traverse[A,B](as: Seq[A])(f: A => Future[B]): Future[Seq[B]]
```

then this function becomes

```scala
def sequenceEffects[A](fs: Seq[() => Future[A]]): Future[Seq[A]] = traverse(fs)(_())
```

mosesn:
What is an example of why I would want to use this?

luciferous:
+1 to @mosesn's question.

Chandler:
The use case I had was trying to rate limit out going requests. My service needed to make thousands of backend requests and I needed to rate limit it dramatically so it was easy to model it by processing each request in sequence and composing a future.sleep onto each request.

If there were a future based, time based rate limiter in util I probably could have used that instead.

mosesn:
@Chandler I'm working on that this week, so that might be able to fit that use case.  Are there any other use cases for this method?

k4rthik:
Hi @mosesn , I needed this last week when I was working on an api that  insert/updates multiple mysql records in a transaction. 
def insertRecord(record: Record): Future[Boolean] = { //do a fixed number of writes in a trasaction like 
for { a <- writeA
              b <- writeB ... })
now I needed another method
def  insertRecords(records: Seq[Record]): Future[Boolean]
that should insert either all or none. Cannot fire them in parallel as they they are all in a single trasaction. 

Glad to hear that you are working on a ratelimiter.

luciferous:
@k4rthik I'm not sure I follow your example. If one operation fails, the operations that were successful up to that one won't roll back. I hope that's not what you were depending on.

k4rthik:
@luciferous Edited the example to be more clear. This should rollback the operations that were successful. Trascation is started in insertRecords and used in each insertRecord.

mosesn:
@k4rthik, I think that @luciferous is right, that you should be able to use the asynchronous insert.  You can only do a transaction over a single session using mysql, and the transactional client gives you exactly one session.  Mysql is over TCP, so we're going to insert all of your records in order (and in this case you don't care about the order that they're inserted in, as far as I can tell).

You should be able to join all of the insertions you've done, and provide that as the return value for the transaction.  If any of them has failed, then it will roll back the transaction.

Chandler:
So what do we think, should we ditch this? @stuhood @dschobel?

To me it seems like a valid way to structure requests if you understand what you're doing (I'm pretty sure my use case is defensible, maybe there are others I haven't thought of).

I like Daniel's traverse/sequenceEffects idea.

On the other hand, this may tempt people into sequencing things that don't need to be sequenced.

mosesn:
@Chandler I think we should make sure that this is really the right abstraction.  I agree that it could be useful, but if comes down to solving the same problem as other tools we already have (like an async rate limiter) then we should prefer not to provide redundant functionality.

With that said, @stuhood said he had seen a bunch of requests for this recently, so my guess is that I just don't understand the use case yet.

dschobel:
@Chandler I think it's fundamentally useful so I'm +1.

As for confusing people, I think the intuition for non-experts with Seq[A]'s that they want to turn into Future[Seq[B]] would be to #map and then look for a method with #collect's signature (but I might just be rationalizing things :) ).

leftparen:
I came across a similar need @Chandler; in my case, sequencing a series of API calls so objects are created in the right order.

mosesn:
@leftparen could you elaborate on your use case?  trying to get a better sense of what this is for.

leftparen:
For example, let's say you have an API that takes one post at a time; you want to send it multiple posts, but you want to make sure the order of posts is preserved. You would need then to wait for the first API response before moving on to the next call.

mosesn:
OK, so it seems like the key thing here is that we generally assume that the reason why you aren't doing something in parallel is because there's a data dependency–in this case, the "data" that you care about is implicit, because all you care about is an acknowledgment that it has processed the write (ie the Future is satisfied).

In that case this API makes sense.  I'd be happy to ship this as is then.

:+1:

luciferous:
@leftparen how are your calls represented? If they are `Seq[Request]`, could this work?

```scala
def doPosts(reqs: Seq[Request]): Future[Unit] =
  reqs match {
    case Nil => Future.Done
    case req +: more => client(req).unit.flatMap(doPosts(more))
  }
```

taylorleese:
@mosesn I'd prefer to ship this with @dschobel's suggestion of traverse/sequenceEffects. Calling this ```sequence``` would be confusing to me and likely others if they are familiar with what sequence means in the scala standard library.

mosesn:
:+1: Sure–seems reasonable.

util-core/src/main/scala/com/twitter/util/Future.scala
johnynek: +1 to not using the name sequence. What about `sequentially`?

Chandler:
taking @dschobel's advice I updated it to be:

traverseSequentially() 
sequenceEffects()

I would have called it only traverse() but in Scala Futures, traverse is not sequential:
https://github.com/scala/scala/blob/2.12.x/src/library/scala/concurrent/Future.scala#L766
"
   *  This is useful for performing a parallel map. For example, to apply a function to all items of a list
   *  in parallel"

johnynek:
Really it might be nice to have a type def for `() => Future[T]` or even a `case class FutureThunk[T](fn: () => Future[T])` since note that is still a Monad (you could write `flatMap` and `sequence` directly on this type (`FutureThunk[T]`).

Chandler:
Maybe we don't need sequenceEffects() at all. To use sequenceEffects you need a function to turn your Seq[Thing] into Seq[FutureThunk].  

And if you have that, you might as well just use the traverse directly like, traverseSequentially(yourThings)(thingToFuture).

Instead of the extra step of building the thunks and then passing them in.

luciferous:
I'm still unsure what's preventing you from modeling what you want with a `Stream` (as `Seq[() => Future[T]]` is effectively `Stream[Future[T]]`) and then use `AsyncStream.fromSeq(stream).toSeq()`, which will give you the properties you want:

1. Sequential traversal.
2. First error abandons traversal and returns it in Future.exception.

Using your rate limiting with `Future.sleep` example:

```scala
def doReq(req: Req): Future[Unit] = send(req) before Future.sleep(1.second)
val reqs: Seq[Req]
val actions: Stream[Future[Unit]] = reqs.toStream.map(doReq)
AsyncStream.fromSeq(actions).toSeq() onFailure { cause =>
  // uh-oh
}
```

util-core/src/main/scala/com/twitter/util/Future.scala
```diff
@@ -339,6 +339,45 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
   def join[A](fs: JList[Future[A]]): Future[Unit] = Futures.join(fs)
 
   /**
+   * Take a sequence and sequentally apply a function f to each item.
```
taylorleese: sequentally => sequentially

util-core/src/main/scala/com/twitter/util/Future.scala
```diff
@@ -339,6 +339,45 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
...
+   * Then return all future results as a single Future[Seq[_]].
+   * 
+   * If during execution any f() throws an exception that exception will be returned and 
+   * the remaining futures will not be processed.
+   * 
+   * @param as a sequence of A that will have f applied to each item sequentially
+   * @return a `Future[Seq[B]]` containing the results of f being applied to every item in as
+   */
+  def traverseSequentally[A,B](as: Seq[A])(f: A => Future[B]): Future[Seq[B]] =
```
taylorleese: traverseSequentally => traverseSequentially

mosesn:
Sort of a sidenote, but `FutureThunk` might be generally useful.  Folks often ask how to use `RetryFilter[(), Rep]` for something asynchronous that doesn't look like a `Service[Req, Rep]`, and right now, the answer is to make a trivial `Service[Unit, Rep]` but we don't really need that.

Chandler:
Anyone who thinks "how can I sequence some futures" is probably going to first look in Futures.scala and then second, write it themselves. It's great that AsyncStream can do this but how is anyone going to know to turn to AsyncStream to accomplish this? Are we going to add that AsyncStream example into the comments of Futures.scala ?

I feel like this is a pretty basic operation that should live next to collect/join etc so it can be used without having to stop and understand what an AsyncStream is.

luciferous:
@Chandler You're absolutely right. The Future object is probably the first place people start looking for functions related to Futures, which happens to be why we keep putting things in there. (And you can see how the Future object grows because of this.)

It's slightly unfortunate that AsyncStream isn't one of the first places to look, because AsyncStream is actually more fundamental than Future. It's easy to see that AsyncStream's dependence on Future is really nothing more than a dependence on the monad interface. We only specialized it to Future for convenience.

What do you think of these?
```
def sequence[A](as: Seq[A]): Future[Seq[A]] =
  AsyncStream.fromSeq(as).toSeq()

def traverseSequentially[A,B](as: Seq[A])(f: A => Future[B]): Future[Seq[B]] =
  AsyncStream.fromSeq(as).mapF(f).toSeq()
```

dschobel:
Given that the problem is defined in Futures and the simplicity of the Future native implementation I don't see the benefit of bringing in an additional abstraction.

util-core/src/main/scala/com/twitter/util/Future.scala
```diff
@@ -339,6 +339,27 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
...
+   * the remaining futures will not be processed.
+   * 
+   * usage:
+   *   Future.traverseSequentially(Seq(1,2,3))(deleteItem)
+   * 
+   * @param as a sequence of A that will have f applied to each item sequentially
+   * @return a `Future[Seq[B]]` containing the results of f being applied to every item in as
+   */
+  def traverseSequentially[A,B](as: Seq[A])(f: A => Future[B]): Future[Seq[B]] =
+    as.foldLeft(Future(Vector.empty[B])) { (resultsFuture, nextItem) =>
```
dschobel: let's use `Future.value` since `Vector.empty` won't throw outside of pathological cases were our vm is already doomed.

util-core/src/main/scala/com/twitter/util/Future.scala
```diff
@@ -339,6 +339,27 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)) map { _ => (%s) }""".format(
   def join[A](fs: JList[Future[A]]): Future[Unit] = Futures.join(fs)
 
   /**
+   * Take a sequence and sequentially apply a function f to each item.
```
dschobel: backticks on `f` please

util-core/src/main/scala/com/twitter/util/Future.scala
luciferous: Future.Nil can work too and saves an allocation.

dschobel: I think the best we can do is stash a singleton Future.value(Vector.empty) in the companion

luciferous: :+1: works for me

Chandler: so like this?
 private val FutureEmptyVector: Future[Vector[Nothing]] = Future.value(Vector.empty)
 ...
 as.foldLeft(FutureEmptyVector.asInstanceOf[Future[Vector[B]]]) { (resultsFuture, nextItem) =>

dschobel: I'm not convinced that the shaved allocation is worth any extra lines of code so I'm still happy with the inline Vector.empty, but your call. If you go the shaving route, we should be able to drop the cast by adding the type param to the foldleft though.

luciferous:
Note: the use of foldLeft will left associate the future flatMaps which incurs ~100b (iirc) per flatMap more than if right associated.

```
Seq(a, b, ...).foldLeft(Future.Nil)(f)
= f(f(Future.Nil, a), b)
= Future.Nil.flatMap(...a...).flatMap(...b...)
```

Right associated flatMaps, e.g.:

```
Future.Nil.flatMap(...a...flatMap(...b...))
```

avoid the ~100b overhead. If this is important have a look at AsyncStream's `observe`.

Chandler:
This is merged internally! @luciferous I'm gonna work in your perf improvement in a new branch.

luciferous:
Good work :) OK.

PR #145, opened 2015-11-05
Fix style error in README

jakequist:
Make the Hashing section look like this:

image

Instead of the current: 

image


mosesn:
Good catch!  LGTM

kevinoliver:
thanks! pulling this in locally.

kevinoliver:
This just got merged in locally and should appear on `develop` this Monday. Thanks again.

PR #146, opened 2015-11-10
Simplify internals and reduce allocations in AsyncStream

luciferous:
AsyncStream is commonly used to represent multiple calls to a source of
Futures as a logical stream, e.g.:

    def source(): Future[A]

    def makeStream(): AsyncStream[A] =
      AsyncStream.fromFuture(source()) ++ makeStream()

AsyncStream.fromFuture maps the future once, and ++ maps it
again. Each call to Future.map creates a new Promise.

    fromFuture(fa) ++ makeStream()
        = fa.map(One(_)) ++ makeStream()           (fromFuture)
        = fa.map(One(_)).map {                     (++)
            case One(a) => Cons(a, makeStream())   (pattern match)
          }

Solution

We can reduce allocations by recognizing that the plain untouched future
is a sufficient encoding for a stream of one element.

    private case class FromFuture[A](fa: Future[A]) extends AsyncStream[A]

    def fromFuture[A](fa: Future[A]) = FromFuture(fa)

It follows from this that a pair of future and an AsyncStream is also a
sufficient encoding for an AsyncStream.

    private case class Cons[A](fa: Future[A], more: () => AsyncStream[A])
      extends AsyncStream[A]

These constructors allow us to defer mapping the future, incurring the
associated allocation costs. Furthermore, generating and consuming
operations sequence futures directly without mapping them through an
intermediate representation. As a bonus, we can do away with LazySeq,
thereby simplifying internals.

    fromFuture(fa) ++ makeStream()
        = FromFuture(fa) ++ makeStream()           (fromFuture)
        = Cons(fa, makeStream())                   (++)

Benchmark

```
(Before)
asyncStream                        20  avgt   10  10615.547 ±  633.728   ns/op
asyncStream:·gc.alloc.rate.norm    20  avgt   10  17272.018 ±    0.061    B/op

(After)
asyncStream                        20  avgt   10   5595.163 ± 101.224   ns/op
asyncStream:·gc.alloc.rate.norm    20  avgt   10  10048.009 ±   0.032    B/op
```

### DIFF
--- a/util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
+++ b/util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
@@ -4,30 +4,6 @@ import com.twitter.util.{Future, Return, Throw}
 import scala.collection.mutable
 
 /**
- * A cons list (a linked list like node) with an AsyncStream tail.
- */
-private sealed trait LazySeq[+A]
-
-private object LazySeq {
-  case object Empty extends LazySeq[Nothing]
-
-  // It'd be nice to make this a value class, but we run into a type erasure
-  // bug: https://issues.scala-lang.org/browse/SI-6260, which seems to be fixed
-  // in 2.11, so let's revisit this in the future.
-  case class One[A](a: A) extends LazySeq[A]
-
-  final class Cons[A] private (val head: A, next: => AsyncStream[A]) extends LazySeq[A] {
-    lazy val tail: AsyncStream[A] = next
-    override def toString = s"Cons($head, ?)"
-  }
-
-  object Cons {
-    def apply[A](head: A, next: => AsyncStream[A]): Cons[A] =
-      new Cons(head, next)
-  }
-}
-
-/**
  * A representation of a lazy (and possibly infinite) sequence of asynchronous
  * values. We provide combinators for non-blocking computation over the sequence
  * of values.
@@ -51,33 +27,35 @@ private object LazySeq {
  *
  * The stream is persistent and can be shared safely by multiple threads.
  */
-final class AsyncStream[+A] private (private val underlying: Future[LazySeq[A]]) {
+abstract class AsyncStream[+A] {
   import AsyncStream._
-  import LazySeq.{Cons, Empty, One}
 
   /**
    * Returns true if there are no elements in the stream.
    */
-  def isEmpty: Future[Boolean] = underlying.flatMap {
+  def isEmpty: Future[Boolean] = this match {
     case Empty => Future.True
+    case M(fas) => fas.flatMap(_.isEmpty)
     case _ => Future.False
   }
 
   /**
    * Returns the head of this stream if not empty.
    */
-  def head: Future[Option[A]] = underlying.map {
-    case Empty => None
-    case One(a) => Some(a)
-    case cons: Cons[A] => Some(cons.head)
+  def head: Future[Option[A]] = this match {
+    case Empty => Future.None
+    case FromFuture(fa) => fa.map(Some(_))
+    case Cons(fa, _) => fa.map(Some(_))
+    case M(fas) => fas.flatMap(_.head)
   }
 
   /**
    * Note: forces the first element of the tail.
    */
-  def tail: Future[Option[AsyncStream[A]]] = underlying.map {
-    case Empty | One(_) => None
-    case cons: Cons[A] => Some(cons.tail)
+  def tail: Future[Option[AsyncStream[A]]] = this match {
+    case Empty | FromFuture(_) => Future.None
+    case Cons(_, more) => Future.value(Some(more()))
+    case M(fas) => fas.flatMap(_.tail)
   }
 
   /**
@@ -89,10 +67,11 @@ final class AsyncStream[+A] private (private val underlying: Future[LazySeq[A]])
    * (a +:: m).uncons == Future.value(Some(a, () => m))
    * }}}
    */
-  def uncons: Future[Option[(A, () => AsyncStream[A])]] = underlying.flatMap {
+  def uncons: Future[Option[(A, () => AsyncStream[A])]] = this match {
     case Empty => Future.None
-    case One(a) => Future.value(Some(a -> (() => empty)))
-    case cons: Cons[A] => Future.value(Some(cons.head -> (() => cons.tail)))
+    case FromFuture(fa) => fa.map(a => Some((a, () => empty)))
+    case Cons(fa, more) => fa.map(a => Some((a, more)))
+    case M(fas) => fas.flatMap(_.uncons)
   }
 
   /**
@@ -121,12 +100,16 @@ final class AsyncStream[+A] private (private val underlying: Future[LazySeq[A]])
    *   AsyncStream(1, 2, 3).takeWhile(_ < 0) = AsyncStream.empty
    * }}}
    */
-  def takeWhile(p: A => Boolean): AsyncStream[A] = AsyncStream(
-    foldRight(empty[A].underlying) { (a, as) =>
-      if (p(a)) Future.value(Cons(a, AsyncStream(as)))
-      else empty.underlying
+  def takeWhile(p: A => Boolean): AsyncStream[A] =
+    this match {
+      case Empty => empty
+      case FromFuture(fa) => M(fa.map(a => if (p(a)) this else empty))
+      case Cons(fa, more) => M(fa.map { a =>
+        if (p(a)) Cons(fa, () => more().takeWhile(p))
+        else more().takeWhile(p)
+      })
+      case M(fas) => M(fas.map(_.takeWhile(p)))
     }
-  )
 
   /**
    * Given a predicate `p` returns the suffix remaining after `takeWhile(p)`:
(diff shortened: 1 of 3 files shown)


luciferous:
I dare anyone to come up with a name better than M.

luciferous:
M

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
```diff
@@ -465,7 +469,11 @@ final class AsyncStream[+A] private (private val underlying: Future[LazySeq[A]])
 }
 
 object AsyncStream {
-  import LazySeq.{Cons, Empty, One}
+  private case object Empty extends AsyncStream[Nothing]
+  private case class M[A](fas: Future[AsyncStream[A]]) extends AsyncStream[A]
```
vkostyukov: How about `Embed`? I don't have a good theory why it should be `Embed`, but it sounds kinda good here.

vkostyukov:
It looks great! :+1:

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
luciferous: Ok I like this. M for embed.

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
```diff
@@ -121,12 +100,16 @@ final class AsyncStream[+A] private (private val underlying: Future[LazySeq[A]])
...
    * }}}
    */
-  def takeWhile(p: A => Boolean): AsyncStream[A] = AsyncStream(
-    foldRight(empty[A].underlying) { (a, as) =>
-      if (p(a)) Future.value(Cons(a, AsyncStream(as)))
-      else empty.underlying
+  def takeWhile(p: A => Boolean): AsyncStream[A] =
+    this match {
+      case Empty => empty
+      case FromFuture(fa) => M(fa.map(a => if (p(a)) this else empty))
```
dschobel: nit: please swap the parens for curlies in map's closure M(fa.map(a => if (p(a)) this else empty))

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
```diff
@@ -137,11 +120,16 @@ final class AsyncStream[+A] private (private val underlying: Future[LazySeq[A]])
    *   AsyncStream(1, 2, 3).dropWhile(_ < 0) = AsyncStream(1, 2, 3)
    * }}}
    */
-  def dropWhile(p: A => Boolean): AsyncStream[A] = AsyncStream(
-    foldRight(empty[A].underlying) { (a, as) =>
-      if (p(a)) as else Future.value(Cons(a, AsyncStream(as)))
+  def dropWhile(p: A => Boolean): AsyncStream[A] =
+    this match {
+      case Empty => empty
+      case FromFuture(fa) => M(fa.map(a => if (p(a)) empty else this))
```
dschobel: same

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
```diff
@@ -196,15 +188,18 @@ final class AsyncStream[+A] private (private val underlying: Future[LazySeq[A]])
...
-      case Empty => empty.underlying
-      case one@One(a) => if (p(a)) Future.value(one) else empty.underlying
-      case cons: Cons[A] =>
-        if (p(cons.head)) Future.value(Cons(cons.head, cons.tail.filter(p)))
-        else cons.tail.filter(p).underlying
+  def filter(p: A => Boolean): AsyncStream[A] =
+    this match {
+      case Empty => empty
+      case FromFuture(fa) =>
+        M(fa.map(a => if (p(a)) this else empty))
```
dschobel: ditto

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
luciferous: You mean like M(fa.map { a => if (p(a)) this else empty })? Oh I see, it could help differentiate the M parens and the map closure...

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
luciferous: So if we look at the type signature `Future[AsyncStream[A]] => AsyncStream[A]`, it's really a lift (`Future[A] => AsyncStream[A]`) and then a flatten (`AsyncStream[AsyncStream[A]] => AsyncStream[A]`).

This doesn't mean I'm suggesting we call this `flatLift`.

util-core/src/test/scala/com/twitter/concurrent/AsyncStreamTest.scala
```diff
@@ -426,4 +426,12 @@ private object AsyncStreamTest {
   def undefined[A]: A = throw new Exception
 
   def toSeq[A](s: AsyncStream[A]): Seq[A] = await(s.toSeq())
+
+  def fromSeq[A](s: Seq[A]): AsyncStream[A] =
```
dschobel: why not use #fromSeq in AsyncStream's companion object?

dschobel:
this is really nice, thank you. one question but otherwise lgtm!

util-core/src/test/scala/com/twitter/concurrent/AsyncStreamTest.scala
luciferous: I wanted to make sure the tests covered all the constructors.

dschobel: makes sense, mind adding a code comment saying as much?

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
```diff
@@ -465,7 +469,11 @@ final class AsyncStream[+A] private (private val underlying: Future[LazySeq[A]])
 }
 
 object AsyncStream {
-  import LazySeq.{Cons, Empty, One}
+  private case object Empty extends AsyncStream[Nothing]
+  private case class Embed[A](fas: Future[AsyncStream[A]]) extends AsyncStream[A]
+  private case class FromFuture[A](fa: Future[A]) extends AsyncStream[A]
+  private case class Cons[A](fa: Future[A], more: () => AsyncStream[A])
```
kevinoliver: want to add back the nicer `Cons.toString` from before?
```
override def toString = s"Cons($fa, ?)"
```

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
```diff
@@ -279,13 +270,15 @@ final class AsyncStream[+A] private (private val underlying: Future[LazySeq[A]])
...
    * }}}
    */
-  def scanLeft[B](z: B)(f: (B, A) => B): AsyncStream[B] = z +:: AsyncStream(
-    underlying.flatMap {
-      case Empty => empty[B].underlying
-      case One(a) => of(f(z, a)).underlying
-      case cons: Cons[A] => cons.tail.scanLeft(f(z, cons.head))(f).underlying
+  def scanLeft[B](z: B)(f: (B, A) => B): AsyncStream[B] =
+    this match {
+      case Embed(fas) => Embed(fas.map(_.scanLeft(z)(f)))
```
kevinoliver: feel free to ignore, but the other methods are ordered Empty, FromFuture, Cons, Embed. the different order made it a touch harder to read. no big deal obviously.

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
luciferous: I overrode it initially because the tail argument was call by name, which toString would have forced, evaluating it prematurely. Now that it's a function, toString won't force it. We could change it back to preserve the '?' but it might be confusing given that argument is now a function.

kevinoliver:
really nice improvements.

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
luciferous: Ah, I put embed first because that's the only time we don't want to prefix with z. What do you think if I added a comment about that and left it in that first position? Open to other formatting ideas too.

kevinoliver: nah, that makes sense to me. just ignore.

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
kevinoliver: makes sense to me.

util-core/src/test/scala/com/twitter/concurrent/AsyncStreamTest.scala
luciferous: Done. Github isn't collapsing this part somehow, maybe because the diff didn't change where you marked it.

taylorleese:
:+1:

luciferous:
Closed by 6629326ef39becb068347e495925d16fb25c69f1.

PR #151, opened 2016-03-08
Close a security hole!

jacksonic:
https://commons.apache.org/proper/commons-collections/security-reports.html


kevinoliver:
thanks @jacksonic, good catch. will merge this in.

kevinoliver:
This got merged in locally and it should appear on the `develop` branch this coming Monday. Thanks for the patch.

mosesn:
Hi @jacksonic, there's good news!  We don't use the vulnerable part of apache commons (the deserialization part) so older versions of util are actually still safe.  Thanks for the PR!

PR #154, opened 2016-03-16
Stop takeWhile and dropWhile at predicate match

luciferous:
Problem

`takeWhile` and `dropWhile` behave like `filter`. In the former, the
result contains all items matching the condition. In the latter, the
result contains all items *not* matching the condition.

Solution

Modify `takeWhile` to discard everything after the first match. For
`dropWhile`, stop checking the items against the condition after the
first match.

Result

    val seq = Seq(1,2,3,2)

    // Before
    fromSeq(seq).takeWhile(_ != 2)    // Seq(1,3)
    fromSeq(seq).dropWhile(_ != 2)    // Seq(2,2)

    // After
    fromSeq(seq).takeWhile(_ != 2)    // Seq(1)
    fromSeq(seq).dropWhile(_ != 2)    // Seq(2,3,2)

### DIFF
--- a/util-core/src/test/scala/com/twitter/concurrent/AsyncStreamTest.scala
+++ b/util-core/src/test/scala/com/twitter/concurrent/AsyncStreamTest.scala
@@ -322,14 +322,14 @@ class AsyncStreamTest extends FunSuite with GeneratorDrivenPropertyChecks {
   }
 
   test("takeWhile") {
-    forAll { (as: List[Int], p: Int => Boolean) =>
-      assert(toSeq(fromSeq(as).takeWhile(p)) == as.takeWhile(p))
+    forAll(genListAndSentinel) { case (as, x) =>
+      assert(toSeq(fromSeq(as).takeWhile(_ != x)) == as.takeWhile(_ != x))
     }
   }
 
   test("dropWhile") {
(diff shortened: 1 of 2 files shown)


util-core/src/test/scala/com/twitter/concurrent/AsyncStreamTest.scala
```diff
@@ -322,14 +322,14 @@ class AsyncStreamTest extends FunSuite with GeneratorDrivenPropertyChecks {
   }
 
   test("takeWhile") {
-    forAll { (as: List[Int], p: Int => Boolean) =>
```
mosesn: for folks who are curious about why this didn't work before, scalacheck generated functions were constant.  we should be able to use scalacheck generated functions if we upgrade to 0.13.0.

https://github.com/rickynils/scalacheck/pull/171

mosesn:
LGTM

kevinoliver:
good catch @luciferous, thanks!

luciferous:
@kevinoliver relied too much on function generation magic :sweat_smile:

mosesn:
@luciferous merged internally!  💖

mosesn:
Made it internally!  https://github.com/twitter/util/commit/adb0aad23731b7fbfa8198cfb80485b901148b33

PR #156, opened 2016-03-20
Preserve Local settings when logging asynchronously.

dimatkach:
Problem

When using QueueingHandler, we are losing the context settings set via Local,
such as TraceId, ClientId, etc., because the Local settings aren't transferred to
the formatting/logging task.

Solution

Use Local.let{} around the formatiing task to let it execute in the context of the
thread that was doing the logging.

Result

When using ansyc logging, the log messages instead of something like this:
  I 0320 14:52:53.600 THREAD46: EchoRequest begin
will now incude full context
  I 0320 14:52:26.006 THREAD46: TraceId:d238dc5aeeb46975 EchoRequest begin

(note, there is also inferClassNames parameter to QueueingHandler, that controls
including of class and method names into the logging messages. This stays unchanged).

### DIFF
--- a/util-logging/src/main/scala/com/twitter/logging/QueueingHandler.scala
+++ b/util-logging/src/main/scala/com/twitter/logging/QueueingHandler.scala
@@ -81,6 +81,8 @@ class QueueingHandler(handler: Handler, val maxQueueSize: Int, inferClassNames: 
 
   import QueueingHandler._
 
+  case class RecordWithLocals(record: javalog.LogRecord, locals: Local.Context)
+
   def this(handler: Handler, maxQueueSize: Int) =
     this(handler, maxQueueSize, false)
 
@@ -90,7 +92,7 @@ class QueueingHandler(handler: Handler, val maxQueueSize: Int, inferClassNames: 
   protected val dropLogNode: String = ""
   protected val log: Logger = Logger(dropLogNode)
 
-  private[this] val queue = new AsyncQueue[javalog.LogRecord](maxQueueSize)
+  private[this] val queue = new AsyncQueue[RecordWithLocals](maxQueueSize)
 
   private[this] val closed = new AtomicBoolean(false)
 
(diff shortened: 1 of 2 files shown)


mosesn:
oh, I understand what you're trying to do now.  could we fix this more simply by changing the implementation of [LazyLogRecord][0]?

[0]: https://github.com/twitter/util/blob/develop/util-logging/src/main/scala/com/twitter/logging/LazyLogRecord.scala#L21-L27

dimatkach:
Well, I wanted it to work for all loggers, not just twitter.util.*.
Besides, I am not sure I understand why you think it would be simpler: we
could save the locals inside LazyLogRecord, but QueuieingHandler would
still need to be modified to restore it around doPublish. Also, this would
affect synchronous case as well, where it is unnecessary.

mosesn:
I think only util-logging uses QueueingHandler, no?  We wouldn't need to modify QueueingHandler because we could restore it in getMessage.  LazyLogRecord isn't used for the synchronous case, so it should be OK, I think.

dimatkach:
util-logging installs QueueingHandler, yes. But third party code is still
doing `LogManager.geLogger` and logging message with jul.Loggers.
I am not sure why you think LazyLogRecord is not used for synchronous case.
It looks like, when there are parameters, it is always used by the
Logger.log:
https://github.com/twitter/util/blob/develop/util-logging/src/main/scala/com/twitter/logging/Logger.scala#L111
and when there aren't, it's only used by the "lazy" calls, but not regular
ones.

dimatkach:
Also, I am not sure how we could to it in `getMessage`: we can set the
saved context in there, but where would the restore happen? Just let it
stay? That seems ... unclean ... especially, since, this is execute on the
default pool.

dimatkach:
And also, twitter.util.Formatter isn't using `getMessage` to format prefix:
https://github.com/twitter/util/blob/develop/util-logging/src/main/scala/com/twitter/logging/Formatter.scala#L189

I know that it's not being used by default (even though, the code looks
confusingly like it would be), but still, I am not sure why we should
prefer a solution that we know does not work in all cases to the one that
does.

What are the problems you see with the approach I am proposing?

mosesn:
OK, let me get this straight, the problem you're running into is that your formatter needs context, not that your actual log needs context?  That makes sense.  I think this is a solid solution then.  LGTM

dimatkach:
Ping ... Is anything happening to this? Just checking ...

mosesn:
Sorry for the delay, I think this is ready to ship.

mosesn:
This was pushed to the develop branch here: https://github.com/twitter/util/commit/50ae991b39e9ec79c147b7df85a818161c750ad6

Thanks for the contribution!

PR #159, opened 2016-04-14
Pass Throwables by-name to Future.raiseWithin

adleong:
Problem:
Throwables passed to Future.raiseWithin are passed by value and are allocated
even if the timeout is never reached.

Solution:
Pass by-name instead.

Result:
The Throwable is only allocated if the timeout is reached.


mosesn:
Heyyyyy, so this is source-compatible for scala but not for java, and it breaks ABI compatibility in a bad way.  Also, call-by-names allocate on every call unless you're using scalac -optimize (which we aren't, often).

Can you cache your exception instead?  The stacktrace is not going to be useful regardless.

adleong:
That's a bummer, but makes sense.  Thanks for the response and explanation!

PR #161, opened 2016-05-03
util-core: toString override for AsyncStream.Cons

luciferous:
Problem

`Cons` changed from a `case class` to `class` and lost the display of its parameters in `toString`. Visibility of these values is useful for debugging.

Solution

Override `toString` to reinstate the `case class` behavior.

Result

Visible head and tail, e.g.:

    scala> println(0 +:: AsyncStream.empty)
    Cons(ConstFuture(Return(0)), <function0>)

### DIFF
--- a/util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
+++ b/util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
@@ -609,6 +609,7 @@ object AsyncStream {
     extends AsyncStream[A] {
     private[this] lazy val _more: AsyncStream[A] = next()
     def more(): AsyncStream[A] = _more
+    override def toString(): String = s"Cons($fa, $next)"
   }
 
   object Cons {


util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
```diff
@@ -609,6 +609,7 @@ object AsyncStream {
     extends AsyncStream[A] {
     private[this] lazy val _more: AsyncStream[A] = next()
     def more(): AsyncStream[A] = _more
+    override def toString(): String = s"Cons($fa, $next)"
```
kevinoliver: am i paranoid to want this to be `${next.toString}`?

kevinoliver:
@luciferous, i <3 better toStrings! have one little request though to help me sleep better.

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
mosesn: yes, the s string interpolator toStrings for you

kevinoliver: ok fair enough. i realize it was doing toString, it just makes me squeamish. not gonna hold this up though.

kevinoliver:
👍

mosesn:
LGTM

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
luciferous: cool. also please don't hesitate to change it in internal review.

`$next` isn't so useful here -- it's always going to be `"<function0>"`. something much more useful is if we made a `Lazy[A]` type. i had a ticket for that idea, you might be able to dig it up. for the `Lazy[A]` type, `toString` would return`"<function0>"` if not yet evaluated, otherwise it would show the evaluated `A`.

luciferous:
Closed by https://github.com/twitter/util/commit/975c9821104e89e6d77c2da36057567451719a99.

Issue #162, opened 2016-05-18
2.12-ification


mosesn:
In order to support 2.12, we need to add crossScalaVersions to Build.scala, and make sure everything compiles.  Not every scala library will be built against the current snapshot, so we may have to support different versions from 2.11 to 2.12.  If part of an API has churned (scalacheck in particular is susceptible to this) we may need to upgrade the 2.11 library version to make progress.  If that happens, contact me (@mosesn) and I'll navigate the upgrade for you–it's a little complicated because we will need to upgrade all of our internal libraries, not just finagle.

I expect most of the modules to be straightforward to upgrade.  The one that might not be is `util-eval`, which we still can't get rid of because `ostrich` depends on it.  I'll see if I can remove it from `ostrich`, but I'm not holding my breath.

- [x] util-app @jeffreyolchovy https://github.com/twitter/util/pull/163
- [x] util-benchmark @jeffreyolchovy 
- [x] util-cache @jeffreyolchovy 
- [x] util-class-preloader @jeffreyolchovy 
- [x] util-codec @jeffreyolchovy 
- [x] util-collection @jeffreyolchovy 
- [x] util-core @jeffreyolchovy 
- [x] util-eval @jeffreyolchovy 
- [x] util-events @jeffreyolchovy 
- [x] util-function @jeffreyolchovy 
- [x] util-hashing @jeffreyolchovy 
- [x] util-jvm @jeffreyolchovy 
- [x] util-lint @jeffreyolchovy 
- [x] util-logging @jeffreyolchovy 
- [x] util-reflect @jeffreyolchovy 
- [x] util-registry @jeffreyolchovy 
- [x] util-stats @jeffreyolchovy 
- [x] util-test @jeffreyolchovy 
- [x] util-thrift @jeffreyolchovy 
- [x] util-zk @jeffreyolchovy 
- [x] util-zk-common @jeffreyolchovy 
- [x] util-zk-test @jeffreyolchovy 

If you want to claim a module, please let us know in this thread!  I'll annotate different modules once they're done.

Probably a good way to start is for someone to try add 2.12 support for all of them, and then report back which ones still fail, so we know which ones are tricky and which ones work trivially.

jeffreyolchovy:
I'll give this one a shot. Will report back with initial results.

jeffreyolchovy:
From a very high level, we're missing 2.12 artifacts for:
- `org.scalacheck#scalacheck`
- `org.scoverage#sbt-scoverage`

[...]

PR #163, opened 2016-05-21
2.12-ification

jeffreyolchovy:
Cross build for Scala 2.12

Problem

There exists a need to publish artifacts that are binary compatible with Scala 2.12.

Solution
- Modify sbt build defintion to cross build for 2.12.
- Modify main and test sources to compile against 2.12.
- Ensure the use of scoverage artifacts that exist
when targeting 2.12.
- Update scalatest and scalacheck libraries,
when necessary, to versions that support 2.12.

Result
- When targeting 2.12, the scoverage artifacts will
still use their 2.11 versions.
- By default, scoverage is disabled when targeting 2.12.
- Extra type information is added to the main and test
sources, where necessary.
- 2.12 compatible versions of `scalatest` and `scalacheck` are used
when targeting 2.12 only.

<hr/>

Corresponding PR to #162 

Currently, all main and test sources are compiling.

~~Still a work in progress as some tests are failing.~~ All tests passing.

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -47,7 +47,7 @@ object Util extends Build {
     unmanagedClasspath in Compile += Attributed.blank(new java.io.File("doesnotexist")),
     libraryDependencies ++= Seq(
       "junit" % "junit" % "4.8.1" % "test",
-      "org.mockito" % "mockito-all" % "1.9.5" % "test",
+      "org.mockito" % "mockito-all" % "1.10.19" % "test",
       scalatestLib(scalaVersion.value) % "test"
     ),
 
@@ -347,7 +347,7 @@ object Util extends Build {
     name := "util-test",
     libraryDependencies ++= Seq(
       scalatestLib(scalaVersion.value),
-      "org.mockito" % "mockito-all" % "1.9.5"
+      "org.mockito" % "mockito-all" % "1.10.19"
     )
   ).dependsOn(utilCore, utilLogging)
 
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -26,12 +26,13 @@ object Util extends Build {
     version := libVersion,
     organization := "com.twitter",
     scalaVersion := "2.11.8",
+    crossScalaVersions := Seq("2.11.8", "2.12.0-M4"),
     // Workaround for a scaladoc bug which causes it to choke on empty classpaths.
     unmanagedClasspath in Compile += Attributed.blank(new java.io.File("doesnotexist")),
     libraryDependencies ++= Seq(
       "junit" % "junit" % "4.8.1" % "test",
       "org.mockito" % "mockito-all" % "1.9.5" % "test",
-      "org.scalatest" %% "scalatest" % "2.2.4" % "test"
+      "org.scalatest" %% "scalatest" % "2.2.6" % "test"
     ),
 
     resolvers += "twitter repo" at "https://maven.twttr.com",
@@ -173,7 +174,7 @@ object Util extends Build {
     name := "util-collection",
     libraryDependencies ++= Seq(
       guavaLib,
-      "org.scalacheck"          %% "scalacheck"          % "1.12.2" % "test"
+      "org.scalacheck" %% "scalacheck" % "1.13.1" % "test"
     )
   ).dependsOn(utilCore % "compile->compile;test->test")
 
@@ -186,7 +187,7 @@ object Util extends Build {
     name := "util-core",
     libraryDependencies ++= Seq(
       "com.twitter.common" % "objectsize" % "0.0.10" % "test",
-      "org.scalacheck" %% "scalacheck" % "1.12.2" % "test",
+      "org.scalacheck" %% "scalacheck" % "1.13.1" % "test",
       "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.4"
     ),
     resourceGenerators in Compile <+=
@@ -254,7 +255,7 @@ object Util extends Build {
     name := "util-hashing",
     libraryDependencies ++= Seq(
       "commons-codec" % "commons-codec" % "1.9" % "test",
-      "org.scalacheck" %% "scalacheck" % "1.12.2" % "test"
+      "org.scalacheck" %% "scalacheck" % "1.13.1" % "test"
     )
   ).dependsOn(utilCore % "test")
 
@@ -302,7 +303,7 @@ object Util extends Build {
       sharedSettings
   ).settings(
     name := "util-stats",
-    libraryDependencies += "org.scalacheck" %% "scalacheck" % "1.12.2" % "test"
+    libraryDependencies += "org.scalacheck" %% "scalacheck" % "1.13.1" % "test"
   ).dependsOn(utilCore, utilLint)
 
   lazy val utilTest = Project(
@@ -313,7 +314,7 @@ object Util extends Build {
   ).settings(
     name := "util-test",
     libraryDependencies ++= Seq(
-      "org.scalatest" %% "scalatest" % "2.2.4",
+      "org.scalatest" %% "scalatest" % "2.2.6",
       "org.mockito" % "mockito-all" % "1.9.5"
     )
   ).dependsOn(utilCore, utilLogging)
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -218,6 +218,7 @@ object Util extends Build {
     name := "util-core",
     libraryDependencies ++= Seq(
       "com.twitter.common" % "objectsize" % "0.0.10" % "test",
+      "org.scala-lang" % "scala-reflect" % scalaVersion.value,
       scalacheckLib(scalaVersion.value) % "test" exclude ("org.scala-lang.modules", "scala-parser-combinators_2.11"),
       "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.4"
     ),
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -22,6 +22,22 @@ object Util extends Build {
   val caffeineLib = "com.github.ben-manes.caffeine" % "caffeine" % "2.3.0"
   val jsr305Lib = "com.google.code.findbugs" % "jsr305" % "2.0.1"
 
+  def scalacheckLib(scalaVersion: String) = {
+    val scalacheckVersion = CrossVersion.partialVersion(scalaVersion) match {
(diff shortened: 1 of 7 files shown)


project/Build.scala
```diff
@@ -172,7 +188,7 @@ object Util extends Build {
     name := "util-collection",
     libraryDependencies ++= Seq(
       guavaLib,
-      "org.scalacheck"          %% "scalacheck"          % "1.12.2" % "test"
+      "org.scalacheck" %% "scalacheck" % "1.13.1" % "test"
```
mosesn: instead of upgrading, can we use a different version between 2.11 and 2.12?  We have better tooling internally to make sure things don't break, but we're still on 1.12 internally.  It'll be easier to make sure things still work for 2.11 (which, frankly, we care a lot more about than 2.12 until it's stable) if we keep the versions aligned with what we use inside of twitter.

project/Build.scala
```diff
@@ -312,7 +328,7 @@ object Util extends Build {
   ).settings(
     name := "util-test",
     libraryDependencies ++= Seq(
-      "org.scalatest" %% "scalatest" % "2.2.4",
+      "org.scalatest" %% "scalatest" % "2.2.6",
```
mosesn: ditto here.

util-core/src/main/scala/com/twitter/util/Bijection.scala
```diff
@@ -57,5 +57,5 @@ trait Bijection[A, B] extends (A => B) { self =>
   /**
    * Composes two instances of Bijection in a new Bijection, with this one applied last.
    */
-  def compose[T](g: Bijection[T, A]) = g andThen this
+  def compose[T](g: Bijection[T, A]): Bijection[T, B] = g andThen this
```
mosesn: we're getting rid of this really soon in exchange for twitter bijections, so let's make sure you merge master before we merge this in.

util-core/src/main/scala/com/twitter/util/Event.scala
```diff
@@ -430,9 +430,7 @@ object Witness {
   /**
    * Create a [[Witness]] from a [[Function1]].
    */
-  def apply[T](f: T => Unit): Witness[T] = new Witness[T] {
-    def notify(t: T): Unit = f(t)
-  }
+  def apply[T](f: T => Unit): Witness[T] = fromFunction(f)
```
mosesn: could you elaborate on what the problem is?  why can't scala do this?  is this a bug in 2.12.0-M4, or is this expected behavior?

mosesn:
Thanks so much for doing this!  One more nit: could you follow the commit message / PR description style described here: https://github.com/twitter/util/blob/develop/CONTRIBUTING.md#style ?  As an example of how it works, you can look at recent commits in the repository.

jeffreyolchovy:
@mosesn, hopefully I've addressed all line notes and comments in latest rebase of full changeset. This is still a wip considering tests are still failing when targeting 2.12 due to "`scalatest` must use a binary compatible version of `scalacheck`". More info available in last commit message and PR description.

project/Build.scala
```diff
@@ -20,21 +20,32 @@ object Util extends Build {
 
   val guavaLib = "com.google.guava" % "guava" % "16.0.1"
 
+  val scalacheckLib = "org.scalacheck" % "scalacheck" % "1.12.2" cross CrossVersion.binaryMapped {
+    case binVersion if binVersion.startsWith("2.12") => "2.11"
+    case binVersion => binVersion
+  }
+
+  def scalatestLib(scalaVersion: String) = {
+    val scalatestVersion = if (scalaVersion.startsWith("2.12")) "2.2.6" else "2.2.4"
```
mosesn: How about using CrossVersion.partialVersion here?

mosesn:
Hmm, it looks like they published scalatest against 1.11.6 [scalacheck][0].  Have you tried using 1.12.5 2.12.0-M4?  That looks [promising][1].

As an aside, it looks like scalatest 3.0 is going to be the way forward [soon][2] so we should start considering upgrading.  However, it isn't going to be a blocker afaict.

[0]: https://repo1.maven.org/maven2/org/scalatest/scalatest_2.12.0-M4/2.2.6/scalatest_2.12.0-M4-2.2.6.pom
[1]: http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22org.scalacheck%22%20AND%20a%3A%22scalacheck_2.12.0-M4%22
[2]: http://www.scalatest.org/release_notes/3.0.0

jeffreyolchovy:
> Hmm, it looks like they published scalatest against 1.11.6 scalacheck. Have you tried using 1.12.5 2.12.0-M4? That looks promising.

That worked, thanks! The `java.lang.IncompatibleClassChangeError` during 2.12 testing has been resolved, but test failures still exist:
```
[error] Failed tests:
[error] 	com.twitter.logging.FormatterTest
[error] Failed tests:
[error] 	com.twitter.util.MonitorTest
[error] 	com.twitter.util.ConfigTest
[error] 	com.twitter.concurrent.OfferTest
```
<hr/>
When targeting 2.12 the `FormatterTest` test-case failure now has a different stack trace than expected:
```
List(
"    at com.twitter.logging.FormatterTest$$.cycle2(FormatterTest.scala:NNN)",
"    at com.twitter.logging.FormatterTest.com$twitter$logging$FormatterTest$$(FormatterTest.scala:NNN)",
"    (...more...)",
"Caused by java.lang.Exception: Aie!",
"    at com.twitter.logging.FormatterTest$$.cycle(FormatterTest.scala:NNN)",
"    at com.twitter.logging.FormatterTest$$.cycle(FormatterTest.scala:NNN)",
"    (...more...)"
)
did not equal
List(
"    at com.twitter.logging.FormatterTest$$.cycle2(FormatterTest.scala:NNN)",
"    at com.twitter.logging.FormatterTest$$.apply$mcV$sp(FormatterTest.scala:NNN)",
"    (...more...)",
"Caused by java.lang.Exception: Aie!",
"    at com.twitter.logging.FormatterTest$$.cycle(FormatterTest.scala:NNN)",
"    at com.twitter.logging.FormatterTest$$.cycle(FormatterTest.scala:NNN)",
"    (...more...)"
) (FormatterTest.scala:181)
```

`ConfigTest`'s test for missing values now shows an additional member:
```
[info]   - should must find top-level and nested missing values *** FAILED ***
[info]     List("bar.z", "com$twitter$util$ConfigTest$$anon$4$$$anonfun$13.z", "x") did not equal List("bar.z", "x") (ConfigTest.scala:74)
[info]   - should must find nested missing values in optional sub-configs *** FAILED ***
[info]     List("baz.w", "com$twitter$util$ConfigTest$$anon$5$$$anonfun$18.w") did not equal List("baz.w") (ConfigTest.scala:85)
```

And at a quick glance, a few Mockito issues makes up the rest of the errors in `OfferTest` and `MonitorTest`. e.g.
```
[info] Cannot call real method on java interface. Interface does not have any implementation!
[info] Calling real methods is only possible when mocking concrete classes.
[info]   //correct example:
[info]   when(mockOfConcreteClass.doStuff()).thenCallRealMethod();))) (of class com.twitter.util.Promise)
[info]     at com.twitter.concurrent.OfferTest.com$twitter$concurrent$OfferTest$$$anonfun$36(OfferTest.scala:228)
[info]     at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.java:12)
```

Will look into these more in depth when I get some time.

mosesn:
@jeffreyolchovy just a heads up, and you don't have to do anything yet, but I'm moving us to scalatest 2.2.6, so we won't have to change the scalatest version based on the version.  I'll see if I can do scalacheck too soon.

jeffreyolchovy:
@mosesn can you review 90199bb which contains a fix for the failing `ConfigTest`? I am using Scala's runtime reflection to ignore synthetic term names when searching for missing values. This eliminates the need for the existing `$outer` suffix check and gets around the (new) issue 2.12 caused as more synthetic methods were being generated in the 'nested' scenarios (e.g. `com$twitter$util$ConfigTest$$anon$5$$$anonfun$18`).

mosesn:
@jeffreyolchovy no worries about the commit message.  for what it's worth, we're going to squash the commits into one, so it's OK as long as just one of the commit messages uses the format.

The new patch LGTM

jeffreyolchovy:
@mosesn, rebased against upstream develop and all tests passing for both 2.11 and 2.12.

util-core/src/test/scala/com/twitter/util/VarTest.scala
```diff
@@ -36,7 +36,7 @@ class VarTest extends FunSuite with GeneratorDrivenPropertyChecks {
     assert(Var.sample(s) == "8923")
 
     var buf = mutable.Buffer[String]()
-    s.changes.register(Witness({ v => buf += v }))
+    s.changes.register(Witness({ (v: String) => buf += v; {} }))
```
mosesn: do you mind using () instead of {}?  doesn't really matter, but it's the style we normally use in util.

mosesn:
LGTM–I also bumped the scalacheck and scalatest versions internally so we don't have to choose them based on the version anymore, but I can fix that when we pull it in. Thanks so much for this!

util-core/src/test/scala/com/twitter/util/VarTest.scala
jeffreyolchovy: np, done.

mosesn:
This was merged internally, and it also now compiles successfully against 2.12.0-M4 in our internal build system!  I'm pretty pumped, thanks for tackling this.  🔥

mosesn:
Merged here: https://github.com/twitter/util/commit/efcc8a3d9933886b271e37a59adc19d37722de83.  With that said, it broke our dependent builds (ostrich / finagle / etc) so we had to revert it, but @cacoco is working on unreverting it.  Thanks!

taylorleese:
@mosesn @cacoco Looks like we need to update the .travis.yml to include 2.12 as well?

cacoco:
@taylorleese I thought we were going to make the build able to be used with 2.12 but until we actually support it we weren't going to guarantee anything, thus *not* update .travis.yml.

taylorleese:
@cacoco That's fine. Let's be consistent with what we do in Ostrich as well though. We suggested updating the .travis.yml there. https://github.com/twitter/ostrich/pull/77

PR #164, opened 2016-05-31
Eval: add Serializable to wrapping class

instanceofme:
Problem

I'm compiling scala code using eval then executing it, in part remotely with Apache Spark. The problem is that if my code contains a class, that class is compiler into an inner class with a non-Serializable outer class (the wrapping `Evaluator__*` class), and it fails to be sent to the remote workers.

Solution

Simply adding `Serializable` to the outer class allows the inner classes to be serialized (as long as they are also `Serializable` obviously). I can execute my compiled code on my remote workers.

Another option would be to make `wrapCodeInClass` (and `codeWrapperLineOffset`) `protected` instead of `private[this]`, so that I can do it myself.

### DIFF
--- a/util-eval/src/main/scala/com/twitter/util/Eval.scala
+++ b/util-eval/src/main/scala/com/twitter/util/Eval.scala
@@ -299,7 +299,7 @@ class Eval(target: Option[File]) {
    * NB: If this method is changed, make sure `codeWrapperLineOffset` is correct.
    */
   private[this] def wrapCodeInClass(className: String, code: String) = {
-    "class " + className + " extends (() => Any) {\n" +
+    "class " + className + " extends (() => Any) with java.io.Serializable {\n" +
     "  def apply() = {\n" +
     code + "\n" +
     "  }\n" +


mosesn:
@instanceofme will this fail if it's not actually Serializable?

For what it's worth, we're trying to end of life `util-eval`, so it might end up disappearing eventually.  This isn't to say we won't take your PR, just wanted to give you a heads up.

instanceofme:
@mosesn Since there is no state and only an `apply()` method, it should be serializable just by implementing `java.io.Serializable`, implementing special serialization handling methods like `writeObject` is not required.

mosesn:
Will the class fail to compile if it has non-serializable members?  Or will it fail at runtime?

instanceofme:
The class cannot have unserializable members, it only has an `apply` method, but can have unserializable inner classes (e.g. in closures). Those will not fail at compile time though, and only fail at runtime if you do serialize them, for instance by sending them in a Spark job. But that's not different than in a regular scala context: you have to make every class that gets sent serializable.
AFAIK there's no downside to making the wrapping class serializable, it will not cause failure in situations that would not fail today, and  it'll make possible the use of serializable inner classes.

mosesn:
@instanceofme rad, works for me.  LGTM

mosesn:
This was pushed to the `develop` branch https://github.com/twitter/util/commit/41e2e9201b595d92a3db5fa1bb74304555715059 thanks for the contribution @instanceofme !

instanceofme:
You're welcome :)

PR #166, opened 2016-06-07
Do not enable scoverage by default

olix0r:
Problem

When running tests against a project that depends on artifacts produced by the current util configuration, the following error is thrown:

    java.io.FileNotFoundException: .../util/util-core/target/scala-2.11/scoverage-data/scoverage.measurements.1147 (No such file or directory)

Solution

Do not set coverage to enabled in Build.scala. Rely on users to manually enable
coverage when it is required (as in travis.yml).

### DIFF
--- a/project/Build.scala
+++ b/project/Build.scala
@@ -113,10 +113,6 @@ object Util extends Build {
         case moduleId =>
           moduleId
       }
-    },
-
-    ScoverageKeys.coverageEnabled := {
-      !scalaVersion.value.startsWith("2.12")
     }
   )
 


taylorleese:
@cacoco ran into this today as well

taylorleese:
Related to https://github.com/twitter/util/pull/163.

cacoco:
@olix0r I tried the changed you're proposing here and that did not seem to work. The original change was reverted here: 8e7e04d70a0b55f06e0177294ba33bd0d5502b29 until we can come up with a proper way forward.

olix0r:
ok weird, i swore that worked for me but maybe i didn't test it properly. thanks for backing it out!

PR #167, opened 2016-06-22
Add scala.util.Try support

johnynek:
Problem

Scala added a `Try` that is basically isomorphic to Twitter's. Code using Twitter Try and scala.util.Try will almost certainly implement methods to convert between these types.

Solution

Add methods to convert to and from these two Trys. 

Result

This will add a few methods which enable these conversions.

### DIFF
--- a/util-core/src/main/scala/com/twitter/util/Future.scala
+++ b/util-core/src/main/scala/com/twitter/util/Future.scala
@@ -104,15 +104,6 @@ object Future {
   def const[A](result: Try[A]): Future[A] = new ConstFuture[A](result)
 
   /**
-   * Creates a satisfied `Future` from a [[scala.util.Try]].
-   *
-   * @see [[value]] for creation from a constant value.
-   * @see [[apply]] for creation from a `Function`.
-   * @see [[exception]] for creation from a `Throwable`.
-   */
-  def const[A](result: scala.util.Try[A]): Future[A] = const(Try.fromScalaTry(result))
-
-  /**
    * Creates a successful satisfied `Future` from the value `a`.
    *
    * @see [[const]] for creation from a [[Try]]
--- a/util-core/src/main/scala/com/twitter/util/Future.scala
+++ b/util-core/src/main/scala/com/twitter/util/Future.scala
@@ -104,6 +104,15 @@ object Future {
   def const[A](result: Try[A]): Future[A] = new ConstFuture[A](result)
 
   /**
+   * Creates a satisfied `Future` from a [[scala.util.Try]].
+   *
+   * @see [[value]] for creation from a constant value.
+   * @see [[apply]] for creation from a `Function`.
+   * @see [[exception]] for creation from a `Throwable`.
+   */
+  def const[A](result: scala.util.Try[A]): Future[A] = const(Try.fromScalaTry(result))
+
+  /**
    * Creates a successful satisfied `Future` from the value `a`.
    *
    * @see [[const]] for creation from a [[Try]]
--- a/util-core/src/main/scala/com/twitter/util/Try.scala
+++ b/util-core/src/main/scala/com/twitter/util/Try.scala
@@ -1,5 +1,7 @@
 package com.twitter.util
 
+import scala.util.{ Success, Failure }
+
 /**
  * The Try type represents a computation that may either result in an exception
  * or return a success value. It is analogous to the Either type but encodes
@@ -14,6 +16,12 @@ object Try {
       case NonFatal(e) => Throw(e)
     }
   }
+
+  def fromScalaTry[R](tr: scala.util.Try[R]): Try[R] =
+    tr match {
+      case Success(r) => Return(r)
+      case Failure(e) => Throw(e)
+    }
 
   /**
    * Like [[Try.apply]] but allows the caller to specify a handler for fatal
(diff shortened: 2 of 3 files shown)


util-core/src/main/scala/com/twitter/util/Try.scala
```diff
@@ -1,5 +1,7 @@
 package com.twitter.util
 
+import scala.util.{ Success, Failure }
```
roanta: style nit: drop the extra spaces around the imports. Alternatively, we can fully qualify these at the usage site to be explicit.

roanta:
lgtm, thanks for the contribution!

dschobel:
lgtm

mosesn:
@johnynek why not encourage folks to use bijections instead?

stuhood:
@mosesn : Given that util can't avoid a dependency on the stdlib anyway, asking people to pull in bijection for this seems like overkill.

johnynek:
I agree with Stu (I don't often say that! ;). This changes no deps and
almost no change to size of the class files. Seems like a strict win.

kevinoliver:
i'm +1 for this change as well.

util-core/src/main/scala/com/twitter/util/Future.scala
```diff
@@ -104,6 +104,15 @@ object Future {
   def const[A](result: Try[A]): Future[A] = new ConstFuture[A](result)
 
   /**
+   * Creates a satisfied `Future` from a [[scala.util.Try]].
+   *
+   * @see [[value]] for creation from a constant value.
+   * @see [[apply]] for creation from a `Function`.
+   * @see [[exception]] for creation from a `Throwable`.
+   */
+  def const[A](result: scala.util.Try[A]): Future[A] = const(Try.fromScalaTry(result))
```
kevinoliver: is this sugar necessary? i'd prefer pointing people at `Try.fromScalaTry` in the `const` scaladocs instead.

util-core/src/main/scala/com/twitter/util/Try.scala
```diff
@@ -15,6 +17,12 @@ object Try {
     }
   }
 
+  def fromScalaTry[R](tr: scala.util.Try[R]): Try[R] =
```
kevinoliver: can you add minimalist scaladoc? 

```
/** Converts from a `scala.util.Try` */
```

mosesn:
@stuhood @kevinoliver @johnynek my feeling here is the same way I feel about Option vs guava Optional / java Optional, which is that if we allow folks to pass in all of the various isomorphisms of these things, we're going to result in unnecessary API duplication when it's pretty simple to simply choose a single type and then allow bijections to move us from the other types here.  In particular, if we had a method which took a `twitter.Try[scala.Option[A]]` (like the result from a cache) we would need to also have a `scala.Try[guava.Optional[A]]`, `scala.Try[java.Optional[A]]`, `scala.Try[scala.Option[A]]`, `twitter.Try[guava.Optional[A]]` etc.

@johnynek any change which increases the surface area of the API isn't a strict win ;)

With that said, I can see the argument that Future is special because it's ubiquitous through our distributed systems code, and it's worth making this easy even if we might not do it everywhere else.  Does that seem fair?

johnynek:
I hear you for sure about API bloat.

That said, `scala.util.Try` is in the standard lib, so I think guava brings us to a whole different set of concerns around adding diamond dependencies. I think the point about something being in the standard library is that everyone will have it.

You guys probably don't see this pain that much because you work in util and maybe very seldom use `scala.util.Try`. As someone who uses util as well as many other libraries that don't use util, it is annoying that every project we have has to either add some bijection import boilerplate or write a few lines of code (two case matches and you are set), when it could just be: `.asScalaTry`.

mosesn:
OK, that makes sense. I think when I reviewed you hadn't made the change to remove it from `Future`, I can see the argument to add this just the `Try` level.  LGTM.

ryanoneill:
+1 for this change also.

util-core/src/main/scala/com/twitter/util/Try.scala
```diff
@@ -67,6 +80,11 @@ object Try {
  */
 sealed abstract class Try[+R] {
   /**
+   * Convert to a scala.util.Try
+   */
+  def asScalaTry: scala.util.Try[R]
```
kevinoliver: (bikeshed, so feel free to drop)

what do you think of shortening this to `asScala` since this going to be something called frequently? (ditto for `fromScala`)

johnynek:
@kevinoliver fine by me to shorten the names.

johnynek:
bummed this did not make it into the release.

Is there anything I can do to get this merged before the next one?

roanta:
@johnynek so sorry, totally forgot to pull this in! I'm putting it up internally now.

johnynek:
please point to this thread if anyone every complains their PRs are
ignored. I worked at twitter for 4.5 years and so are mine!

;P

nothing but <3

On Fri, Jul 8, 2016 at 9:12 AM, Ruben Oanta <notifications@github.com>
wrote:

> @johnynek <https://github.com/johnynek> so sorry, totally forgot to pull
> this in! I'm putting it up internally now.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/twitter/util/pull/167#issuecomment-231447672>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAEJdrDBqcaPwLxdacG7hZDMnvg90f26ks5qTqEAgaJpZM4I8HZH>
> .
>



-- 
P. Oscar Boykin, Ph.D. | http://twitter.com/posco | http://pobox.com/~boykin

roanta:
@johnynek Thanks again for the patch! It's landed here: https://github.com/twitter/util/commit/1b05cf12263d27fe4c390ca5f3d69a7991b30d40

johnynek:
👍

PR #168, opened 2016-08-01
util-eval: Reset reporter between checks

tbekas:
Problem

When you want to use the util-eval for validating the correctness of a small chunk of code, you have to reset whole state between invocations (flag resetState=true), which is expensive when you already have a bunch of compiled an loaded classes. Without this, if one check fails, all subsequent checks will fail as well, since some errors has been reported previously.

Solution

Scala compiler API allows for resetting just the reporter, so that compiled and loaded classes can remain for re-usage for next checks.

Result

Public API remains the same, however the internal API has one new synchronized method `resetReporter()` which does what is explained in the "Solution" section.

### DIFF
--- a/util-eval/src/main/scala/com/twitter/util/Eval.scala
+++ b/util-eval/src/main/scala/com/twitter/util/Eval.scala
@@ -242,6 +242,7 @@ class Eval(target: Option[File]) {
     val id = uniqueId(sourceForString(code))
     val className = "Evaluator__" + id
     val wrappedCode = wrapCodeInClass(className, code)
+    resetReporter()
     compile(wrappedCode) // may throw CompilerException
   }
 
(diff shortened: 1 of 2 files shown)


mosesn:
@tbekas fwiw, we're hoping to kill `util-eval` eventually, but this LGTM, we're still happy to merge this in.

ryanoneill:
Going to attempt to merge this in internally.

ryanoneill:
@tbekas this pull request has been merged internally and should sync back out to Github in a few days. When it does, we'll close this PR. Thanks again.

tbekas:
@mosesn Thanks for the review! 

@ryanoneill Thank you for merging it! Do you know maybe whether there will be a cross build executed, so that there will be available artifacts for both scala 2.10 and 2.11?

ryanoneill:
@tbekas artifacts will be available when we do a release of finagle, which will happen sometime this month (perhaps next week). The code will be available on the develop branch in github by the beginning of next week, at the latest.

Regarding 2.10, our release of Finagle 6.35 in April was the last release planned which targets Scala 2.10. Please see this [blog post](http://finagle.github.io/blog/2016/04/20/scala-210-and-java7/) for more details.

tbekas:
@ryanoneill Thanks for the info and the link, cheers

ryanoneill:
Your commit has been pushed to the develop branch of util https://github.com/twitter/util/commit/9176942591159d3705bd5b3c982cf1c812ac9ca3
I am now going to close this pull request.

Thanks again for the contribution.

PR #170, opened 2016-09-02
reset reporter while apply compiling

nilday:
Problem

I notice that the former release 6.36.0 reset the reporter after check(), which is not enough. I my use case, I would like to use apply with resetState=false (or simply inPlace), without check it before ( check method requires compiling every time, but apply method has cache).
Under such circumstance, if I offer a pice of code with syntax error, it will throw CompilerException every time even if I correct it. It’s because the reporter hasn’t been reset. 

Solution

Simply add resetReporter before compile will solve the problem.

Result

It is self-explanatory based on the solution.

### DIFF
--- a/util-eval/src/main/scala/com/twitter/util/Eval.scala
+++ b/util-eval/src/main/scala/com/twitter/util/Eval.scala
@@ -580,6 +580,9 @@ class Eval(target: Option[File]) {
       if (Debug.enabled)
         Debug.printWithLineNumbers(code)
 
+      //reset reporter, or will always throw exception after one error while resetState==false
+      resetReporter()
+
       // if you're looking for the performance hit, it's 1/2 this line...
       val compiler = new global.Run
       val sourceFiles = List(new BatchSourceFile("(inline)", code))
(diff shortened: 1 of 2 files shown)


mosesn:
Just so you know, we no longer use `util-eval` internally, and it's officially deprecated.  But we're still willing to accept pull requests for it.  Can you add a test?

nilday:
Of course, I will do that. And may I ask what you will use for script eval now?

mosesn:
@nilday we don't compile scala at runtime, we've decided it's not a good practice.  It's difficult to test, it fails at runtime instead of at compile-time, and it's often easy to rewrite these kinds of programs to be compiled instead.  If your use case requires that you be able to evaluate scala at runtime, I think they exist, and they're probably better maintained than `util-eval`.

mosesn:
LGTM

cacoco:
@nilday this has been merged internally and will be pushed to Github on our next sync.

cacoco:
@nilday merged in b0696d02f95ce044761700cd4747254f4bc1bcbc

PR #179, opened 2016-10-21
Allow overriding classloader used by eval. Necessary for classloader hierachies.

cvogt:
Problem

When trying to use Eval in a runtime environment where libraries are isolated in their own classloaders (Scala cbt, assumingly OSGI), Eval.getClass.getClassloader cannot load classes outside of the twitter libraries. The code that is trying to be evaluated may however rely on such classes which currently leads to a ClassNotFound exception.

Solution

I added an override hook for which classloader to use, which the runtime environment can use to inject the correct classloader.

Result

No effects on existing consumers, as they do not override this new member.

### DIFF
--- a/util-eval/src/test/scala/com/twitter/util/EvalTest.scala
+++ b/util-eval/src/test/scala/com/twitter/util/EvalTest.scala
@@ -248,7 +248,7 @@ class EvalTest extends WordSpec {
 
     "eval with overriden classloader" in {
       val eval = new Eval{
-        override def classLoader = this.getClass.getClassLoder
+        override def classLoader = this.getClass.getClassLoader
       }
       assert(eval.apply[Int]("1 + 1") == 2)
     }
--- a/util-eval/src/test/scala/com/twitter/util/EvalTest.scala
+++ b/util-eval/src/test/scala/com/twitter/util/EvalTest.scala
@@ -245,5 +245,12 @@ class EvalTest extends WordSpec {
         assert(eval.errors.isEmpty)
       }
     }
+
+    "eval with overriden classloader" in {
+      val eval = new Eval{
+        override def classLoader = this.getClass.getClassLoder
+      }
+      assert(eval.apply[Int]("1 + 1") == 2)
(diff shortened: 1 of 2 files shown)


ryanoneill:
Hi @cvogt, just so you're aware, we no longer use util-eval internally, and it's officially deprecated. We're still willing to accept a pull request for it however. Would you be able to add a test here? Thanks.

bryce-anderson:
@cvogt, are you still interested in getting this merged?

cvogt:
@bryce-anderson @ryanoneill hej guys, missed the first comment. yeah interested in getting this merged! I added a test.

util-eval/src/test/scala/com/twitter/util/EvalTest.scala
```diff
@@ -245,5 +245,12 @@ class EvalTest extends WordSpec {
         assert(eval.errors.isEmpty)
       }
     }
+
+    "eval with overriden classloader" in {
+      val eval = new Eval{
+        override def classLoader = this.getClass.getClassLoder
+      }
+      assert(eval.apply[Int]("1 + 1") == 2)
```
bryce-anderson: TBH, I'm not super familiar with using different class loaders, but it doesn't appear to me that this test will exercise the logic you've implemented.
Maybe I'm wrong, so just let me know if that's the case.

cvogt:
Since Eval is deprecated and avoiding the dependency on twitter-util is nice in my case, I decided to fork/copy Eval.scala into the CBT code base instead, which is also Apache 2.0 licensed. The PR is linked above. Let me know if there are any concerns about this. I largely preserved history.

I will be able to do the necessary changes for my purposes in my fork, so I am closing this PR.

bryce-anderson:
@cvogt: that seems like a reasonable course of action to me. Sorry this didn't end up getting merged, but sounds like your needs have been sorted out in way that is better in the long term.

cvogt:
@bryce-anderson no worries and thank you :)

PR #183, opened 2016-12-06
removing deprecated QueueProxy in util-core FactoryPool/SimplePool

cevaris:
### Problem

Found a compiler warning, `mutable.QueueProxy` in `Factory/SimplePool` implementation; specifically a helper class `HealthyQueue`.

```
[warn] /git/util/util-core/src/main/scala/com/twitter/util/Pool.scala:60: trait QueueProxy in package mutable is deprecated: Proxying is deprecated due to lack of use and compiler-level support.
[warn]   extends mutable.QueueProxy[Future[A]]
[warn]                   ^
``` 

### Solution

Being that we are using mutable collections, we can modify HealthyQueue to extend `mutable.Queue[Future[A]]` instead of `mutable.QueueProxy[Future[A]]`

### Result
The underlying impl. was always `mutable.Queue[Future[A]]`, so there should be no interface/functionality change.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,6 +8,10 @@ Next Version
 
 6.25.0  2015-06-22
 
+API Changes:
+  * util-core: `FactoryPool`/`SimplePool` now inherits
+  `scala.collection.mutable.Queue[A]` not deprecated `scala.collection.mutable.QueueProxy[A]`
+
 Changes in Runtime Behavior:
   * util-events: Enable event sink by default.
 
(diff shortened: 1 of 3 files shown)


mosesn:
This is a breaking API change, do you mind adding a note to CHANGES?  LGTM otherwise.

kevinoliver:
yep, same thoughts from me.

yufangong:
Thank you @cevaris. This commit has been merged https://github.com/twitter/util/commit/9df5f2b83645cdfab41ecc1c136683435bd97fb4, Thanks again for the contribution.

PR #189, opened 2017-05-09
Modify closeOnExit to call close after app exit

siggy:
Problem

We've run into a race condition where our app is shutdown prior to
closeOnExit getting called. The Closables that were added via
closeOnExit subsequent to app shutdown are never closed.

Solution

Modify closeOnExit to check if shutdown has already been called. If it
is, close the Closable inline.

Result

The caller need not be concerned with causing a race condition between
app.close() and app.closeOnExit().

### DIFF
--- a/util-app/src/main/scala/com/twitter/app/App.scala
+++ b/util-app/src/main/scala/com/twitter/app/App.scala
@@ -104,7 +104,6 @@ trait App extends Closable with CloseAwaitably {
 
   /**
    * This is satisfied when all members of `exits` and `lastExits` have closed.
-   * Note
    *
    * @note Access needs to be mediated via the intrinsic lock.
    */
--- a/util-app/src/main/scala/com/twitter/app/App.scala
+++ b/util-app/src/main/scala/com/twitter/app/App.scala
@@ -105,7 +105,16 @@ trait App extends Closable with CloseAwaitably {
   /**
    * Close `closable` when shutdown is requested. Closables are closed in parallel.
    */
-  final def closeOnExit(closable: Closable): Unit = exits.add(closable)
+  final def closeOnExit(closable: Closable): Unit = {
+    exits.add(closable)
+
+    // We check `closeDeadline` here to ensure that `closable` was added to
+    // `exits` prior to `close()` getting called. If `close()` has already been
+    // called, we should just close `closable` here.
+    if (closeDeadline != Time.Top) {
+      closable.close(closeDeadline)
+    }
+  }
 
   /**
    * Register a `closable` to be closed on application shutdown after those registered
@@ -118,7 +127,13 @@ trait App extends Closable with CloseAwaitably {
    *
    *       In all cases, the close deadline is enforced.
    */
-  final def closeOnExitLast(closable: Closable): Unit = lastExits.add(closable)
+  final def closeOnExitLast(closable: Closable): Unit = {
+    lastExits.add(closable)
+
+    if (closeDeadline != Time.Top) {
+      closable.close(closeDeadline)
+    }
+  }
 
   /**
    * Invoke `f` when shutdown is requested. Exit hooks run in parallel and are
--- a/util-app/src/main/scala/com/twitter/app/App.scala
+++ b/util-app/src/main/scala/com/twitter/app/App.scala
@@ -104,6 +104,9 @@ trait App extends Closable with CloseAwaitably {
 
   /**
    * This is satisfied when all members of `exits` and `lastExits` have closed.
+   * Note
+   *
+   * @note Access needs to be mediated via the intrinsic lock.
    */
   @volatile private[this] var closing: Future[Unit] = Future.never
 
@@ -138,7 +141,9 @@ trait App extends Closable with CloseAwaitably {
     } else {
       // `close()` already called, we need to close this here, but only
       // after `close()` completes and `closing` is satisfied
-      closing.transform { _ => closable.close(closeDeadline) }
+      closing
+        .transform { _ => closable.close(closeDeadline) }
+        .by(shutdownTimer, closeDeadline)
     }
   }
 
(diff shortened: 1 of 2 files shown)


util-app/src/main/scala/com/twitter/app/App.scala
```diff
@@ -118,7 +127,13 @@ trait App extends Closable with CloseAwaitably {
    *
    *       In all cases, the close deadline is enforced.
    */
-  final def closeOnExitLast(closable: Closable): Unit = lastExits.add(closable)
+  final def closeOnExitLast(closable: Closable): Unit = {
+    lastExits.add(closable)
+
+    if (closeDeadline != Time.Top) {
```
mosesn: it's possible for this to create a race condition where this is closed before the "first" closables have started.  maybe we should fix this with mutexes?

util-app/src/main/scala/com/twitter/app/App.scala
```diff
@@ -105,7 +105,16 @@ trait App extends Closable with CloseAwaitably {
...
    * Close `closable` when shutdown is requested. Closables are closed in parallel.
    */
-  final def closeOnExit(closable: Closable): Unit = exits.add(closable)
+  final def closeOnExit(closable: Closable): Unit = {
+    exits.add(closable)
+
+    // We check `closeDeadline` here to ensure that `closable` was added to
+    // `exits` prior to `close()` getting called. If `close()` has already been
+    // called, we should just close `closable` here.
+    if (closeDeadline != Time.Top) {
```
mosesn: uhhhh this is getting really deep in the concurrency weeds but volatile isn't really strong enough here.  the JVM could reorder this in a dirty way:

t1:
```
if (closeDeadline != Time.top)
```

t2: 
```
closeDeadline = Time.now + 5.seconds
closables.foreach(_.close(deadline))
```

t1:
```
exits.add(closable)
```
Pretty good explanation [here][0].

This probably isn't an issue in practice, but I'd rather program against the JMM than against an implementation of the JMM.  Let's just switch this to use mutexes.

[0]: https://shipilev.net/blog/2016/close-encounters-of-jmm-kind/#wishful-hb-actual

util-app/src/main/scala/com/twitter/app/App.scala
```diff
@@ -105,7 +105,16 @@ trait App extends Closable with CloseAwaitably {
   /**
    * Close `closable` when shutdown is requested. Closables are closed in parallel.
    */
-  final def closeOnExit(closable: Closable): Unit = exits.add(closable)
+  final def closeOnExit(closable: Closable): Unit = {
+    exits.add(closable)
```
mosesn: this may cause the closable to closed twice–this is *probably* ok but it's not part of the closable contract that it is, so let's close it exactly once if possible?  again, mutexes seem to be the right tool here.

util-app/src/test/scala/com/twitter/app/AppTest.scala
```diff
@@ -128,4 +128,37 @@ class AppTest extends FunSuite {
       assert(f.isDefined)
     }
   }
+
+  test("App: closes all closables that were added by closeOnExit") {
+    val app = new TestApp(() => ())
+
+    @volatile var closed = false
```
mosesn: nothing is multithreaded here, let's drop the volatile annotation.

util-app/src/test/scala/com/twitter/app/AppTest.scala
```diff
@@ -128,4 +128,37 @@ class AppTest extends FunSuite {
...
+
+    assert(!closed)
+    app.main(Array.empty)
+    assert(closed)
+  }
+
+  test("App: closes all closables that were added by closeOnExit after app exited") {
+    val app = new TestApp(() => ())
+
+    @volatile var closed = false
```
mosesn: same

siggy:
Thanks for the feedback @mosesn ! I've add synchronize blocks to `close()`, `closeOnExit()` and `closeOnExitLast()`.

I needed a way to ensure `closeOnExitLast()` did not close anything before all members of `exits` had closed. I ended up adding a `closing` var that is satisfied when everything else has closed. I don't love adding state like this, so if you see a better way, I'm open to it.

util-app/src/main/scala/com/twitter/app/App.scala
```diff
@@ -118,7 +131,16 @@ trait App extends Closable with CloseAwaitably {
...
    */
-  final def closeOnExitLast(closable: Closable): Unit = lastExits.add(closable)
+  final def closeOnExitLast(closable: Closable): Unit = synchronized {
+    if (closing == Future.never) {
+      // `close()` not yet called, safe to add it
+      lastExits.add(closable)
+    } else {
+      // `close()` already called, we need to close this here, but only
+      // after `close()` completes and `closing` is satisfied
+      closing.transform {_ => closable.close(closeDeadline) }
```
mosesn: should be `{ _` instead of `{_`

siggy:
Thanks for the feedback @dschobel ! I put the deadline enforcement in a slightly different place than where you had suggested, in the `close()` rather than `closeOnExitLast()`. I did this to ensure that `close()` returns a `Future` that should be guaranteed to be eventually satisfied. Let me know what you think.

In addition to adding a test case for second stalled phase, I also modified the `App: sequenced exits respect deadline` test to confirm that the `Future` returned from `close()` is satisfied when a second phase stalls.

util-app/src/main/scala/com/twitter/app/App.scala
```diff
@@ -103,9 +103,25 @@ trait App extends Closable with CloseAwaitably {
   @volatile private[this] var closeDeadline = Time.Top
 
   /**
+   * This is satisfied when all members of `exits` and `lastExits` have closed.
+   * Note
```
dschobel: s/Note// please

siggy: gah my bad!

mosesn:
This was merged in here: https://github.com/twitter/util/commit/93a88444c597841e0a514538b2381b14ddbcf687 thanks for the PR!

PR #191, opened 2017-05-18
Enable ByteReader  to read unsigned longs

mehmetgunturkun:
Problem

readUnsignedLongBE and readUnsignedLongLE are missing in ByteReader implementation. finagle project needed that and as @vkostyukov stated, they're generally usefull, I moved that logic into ByteReader.

Solution

Added two functions 
`readUnsignedLongBE` and `readUnsignedLongLE` to ByteReader and implemented in ByteReaderImpl.

### DIFF
--- a/util-core/src/main/scala/com/twitter/io/ByteReader.scala
+++ b/util-core/src/main/scala/com/twitter/io/ByteReader.scala
@@ -401,25 +401,35 @@ private class ByteReaderImpl(buf: Buf) extends ByteReader {
   private val unsignedLongMaxValue: BigInt = BigInt("18446744073709551615")
 
   def readUnsignedLongBE(): BigInt = {
-    val arr = Array.ofDim[Byte](8)
+    checkRemaining(8)
+    val ret =
+      BigInt(buf.get(pos    ) & 0xff) << 56 |
+      BigInt(buf.get(pos + 1) & 0xff) << 48 |
+      BigInt(buf.get(pos + 2) & 0xff) << 40 |
+      BigInt(buf.get(pos + 3) & 0xff) << 32 |
+      BigInt(buf.get(pos + 4) & 0xff) << 24 |
+      BigInt(buf.get(pos + 5) & 0xff) << 16 |
+      BigInt(buf.get(pos + 6) & 0xff) <<  8 |
+      BigInt(buf.get(pos + 7) & 0xff)
+    pos += 8
 
-    for (i <- 0 until 8) {
-      val b = readByte()
-      arr(i) = b
-    }
-
-    BigInt(arr) & unsignedLongMaxValue
+    ret & unsignedLongMaxValue
   }
 
   def readUnsignedLongLE(): BigInt = {
-    val arr = Array.ofDim[Byte](8)
+    checkRemaining(8)
+    val ret =
+      BigInt(buf.get(pos    ) & 0xff)       |
+      BigInt(buf.get(pos + 1) & 0xff) <<  8 |
+      BigInt(buf.get(pos + 2) & 0xff) << 16 |
+      BigInt(buf.get(pos + 3) & 0xff) << 24 |
+      BigInt(buf.get(pos + 4) & 0xff) << 32 |
+      BigInt(buf.get(pos + 5) & 0xff) << 40 |
+      BigInt(buf.get(pos + 6) & 0xff) << 48 |
+      BigInt(buf.get(pos + 7) & 0xff) << 56
+    pos += 8
 
-    for (i <- 0 until 8) {
-      val b = readByte()
-      arr(7 - i) = b
-    }
-
-    BigInt(arr) & unsignedLongMaxValue
+    ret & unsignedLongMaxValue
   }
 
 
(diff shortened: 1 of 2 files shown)


util-core/src/main/scala/com/twitter/io/ByteReader.scala
```diff
@@ -384,6 +398,41 @@ private class ByteReaderImpl(buf: Buf) extends ByteReader {
     ret
   }
 
+  private val unsignedLongMaxValue: BigInt = BigInt("18446744073709551615")
+
+  def readUnsignedLongBE(): BigInt = {
```
mosesn: this makes a lot of intermediate allocations that I think we can probably skip.  how about we use regular longs except for `BigInt(buf.get(pos    ) & 0xff) << 56` and then or them at the end?

also, why do we need to & it with the unsignedLongMaxValue?

mehmetgunturkun: I fixed unnecessary allocations.

> also, why do we need to & it with the unsignedLongMaxValue?

Frankly, I saw that in existing functions, for instance, _readUnsignedIntLE,_ I guess they're used just in case. Maybe to reset other bits to 0.

util-core/src/main/scala/com/twitter/io/ByteReader.scala
```diff
@@ -419,13 +419,13 @@ private class ByteReaderImpl(buf: Buf) extends ByteReader {
...
     checkRemaining(8)
     val ret =
-      BigInt(buf.get(pos    ) & 0xff)       |
-      BigInt(buf.get(pos + 1) & 0xff) <<  8 |
-      BigInt(buf.get(pos + 2) & 0xff) << 16 |
-      BigInt(buf.get(pos + 3) & 0xff) << 24 |
-      BigInt(buf.get(pos + 4) & 0xff) << 32 |
-      BigInt(buf.get(pos + 5) & 0xff) << 40 |
-      BigInt(buf.get(pos + 6) & 0xff) << 48 |
+            (buf.get(pos    ) & 0xff).toLong       |
```
mosesn: looks like the indentation got screwed up

util-core/src/main/scala/com/twitter/io/ByteReader.scala
```diff
@@ -404,13 +404,13 @@ private class ByteReaderImpl(buf: Buf) extends ByteReader {
...
     val ret =
       BigInt(buf.get(pos    ) & 0xff) << 56 |
-      BigInt(buf.get(pos + 1) & 0xff) << 48 |
-      BigInt(buf.get(pos + 2) & 0xff) << 40 |
-      BigInt(buf.get(pos + 3) & 0xff) << 32 |
-      BigInt(buf.get(pos + 4) & 0xff) << 24 |
-      BigInt(buf.get(pos + 5) & 0xff) << 16 |
-      BigInt(buf.get(pos + 6) & 0xff) <<  8 |
-      BigInt(buf.get(pos + 7) & 0xff)
+            (buf.get(pos + 1) & 0xff).toLong << 48 |
```
mosesn: I think this is still constructing many intermediate objects.  can you only OR with the bigint at the very end?

mehmetgunturkun: Yeah sorry, just send the new push

bryce-anderson:
I believe this will also require a change in finagle-netty4 which contains an implementation of `ByteReader`: https://github.com/twitter/finagle/blob/develop/finagle-netty4/src/main/scala/com/twitter/finagle/netty4/AbstractByteBufByteReader.scala

mosesn:
@bryce-anderson it's probably easiest for us to just do it on behalf of @mehmetgunturkun.  the other thing we could do is merge a finagle and a util commit (which should be possible) but we haven't tried it before.

bryce-anderson:
I'm cool with either, just wanted to note that it would have to be done. 😀

util-core/src/main/scala/com/twitter/io/ByteReader.scala
nepthar: Close. Since java doesn't have any unsigned types, we have to use the next larger type. This facilitates that conversion: https://stackoverflow.com/questions/9578639/best-way-to-convert-a-signed-integer-to-an-unsigned-long. It is not necessary in this case.

Also, BigInt provides a constructor directly from a byte array!

`BigInt(1, Array(buf.get(pos), buf.get(pos+1), buf.get(pos+2)... )`

The first argument is how to interp. So 1 for signed, and -1 for unsigned (I believe)

mehmetgunturkun: > It is not necessary in this case.
Should remove that last & operation if it is unnecessary?

`BigInt(1,` Array(buf.get(pos), buf.get(pos+1), buf.get(pos+2)... )
I know that too but just do not want to allocate another Array.

nepthar: Yeah, remove that last `&`.

SGTM w/avoiding allocations.

util-core/src/main/scala/com/twitter/io/ByteReader.scala
```diff
@@ -384,6 +398,39 @@ private class ByteReaderImpl(buf: Buf) extends ByteReader {
     ret
   }
 
+  def readUnsignedLongBE(): BigInt = {
+    checkRemaining(8)
+    val ret =
+        (buf.get(pos + 7) & 0xff).toLong        |
```
mosesn: please fix the indentation.  two spaces please.

mosesn:
@mehmetgunturkun this was merged in a couple days ago, thanks!  https://github.com/twitter/util/commit/94ab01fc7a568f5bdcb14f032ccb17ed313eb187

PR #192, opened 2017-05-21
Reusing Local.let in closed

ahanwadi:
Problem

Solution

Describe the modifications you've done.

Result

What will change as a result of your pull request? Note that sometimes
this section is unnecessary because it is self-explanatory based on
the solution.


mosesn:
@ahanwadi we don't use `Local#let` here because `Local#let` uses a call-by-name parameter, and so allocates a closure on invocation.  The pattern `local.closed(() => doSomeWork)` is an improvement over `() => local.let(doSomeWork())` mostly because it lets us skip allocating that closure each time.  I'm going to close this PR for now, but please feel free to keep commenting if you have more questions.

PR #197, opened 2017-08-17
util-core: AsyncStream.scanLeft is sometimes one element behind

mejran:
Problem

AsyncStream.scanLeft is one element behind when the original
 AsyncStream is an Embed. The Embed case does not return the initial
 element until the wrapped future is done while other cases return
 the initial value before the Future is done.

Solution

Have the Embed case return the initial value without waiting for the
 future and call a new method, scanLeftEmbed, on the nested AsyncStream.
 scanLeftEmbed which does not return the initial value and thus prevent
 duplicate values from being returned.

Result

scanLeft is now fully eager.


roanta:
Thanks for the patch, this landed: https://github.com/twitter/util/commit/d69429b54b5616df9eb2e4f16dcdc4b49b34a71b

PR #199, opened 2017-08-19
Fixed possible ArrayIndexOutOfBoundException of Logger

Eternity-Yarr:
Problem

It is not safe to assume that stack traces are always available, one might run JVM with `-XX:-StackTraceInThrowable` option.

Solution

I've added boundary check, if there is no stack traces in throwables, simply fall back to some constant name

Result

Finagle will stop failing startup  :)


roanta:
Thanks for the patch, this landed: https://github.com/twitter/util/commit/a4a6b5019cae7a3e2de29272c5d586f9bb007b37

PR #201, opened 2017-09-08
Return Future.never for Future.Sleep(Duration.Top)

adleong:
Problem

Calling Future.Sleep with extremely large values (including Duration.Top) can sometimes return a Future that is satisfied immediately due to overflow in the underlying Timer implementation.

Solution

Since Duration.Top is a special value, we just return Future.never.

Result

Calling Future.sleep with Duration.Top will return a Future that never completes.  Depending on the Timer implementation, calling Future.sleep with other extremely large values can still return a Future that completes immediately.


vkostyukov:
Thanks, @adleong! I'm pulling this internally.

vkostyukov:
Merged in e774ab6082b01cb1ed6f25e409f4ae5c85f89b76. Thanks @adleong!

PR #208, opened 2018-01-15
Wait for resource disposal before closing on error

lilred:
Self-explanatory.

### DIFF
--- a/util-app/src/main/scala/com/twitter/app/App.scala
+++ b/util-app/src/main/scala/com/twitter/app/App.scala
@@ -63,7 +63,7 @@ trait App extends Closable with CloseAwaitably {
 
   protected def exitOnError(reason: String): Unit = {
     System.err.println(reason)
-    close()
+    Await.result(close(), closeDeadline - Time.now)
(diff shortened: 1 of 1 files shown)


util-app/src/main/scala/com/twitter/app/App.scala
```diff
@@ -63,7 +63,7 @@ trait App extends Closable with CloseAwaitably {
 
   protected def exitOnError(reason: String): Unit = {
     System.err.println(reason)
-    close()
+    Await.result(close(), closeDeadline - Time.now)
```
kevinoliver: i think we should handle the failed `Future` case, otherwise the `System.exit` won't get called. also, i think something like:

`Await.result(close().unit, closeDeadline - Time.now)`

ryanoneill:
Hi @lilred, thanks for the contribution. This pull request has been reviewed and merged internally and pushed back out to GitHub as https://github.com/twitter/util/commit/aa32eb16f03635dbdc9540366a880d648aa58949

Issue #214, opened 2018-05-04
ScalaCheck 1.14.0 upgrade fails compilation


SethTisue:
I am upgrading the Scala 2.12 community build to use ScalaCheck 1.14.0 (https://github.com/scala/community-builds/pull/718)

doing so uncovered the following source incompatibility, reproducible outside dbuild by bumping the ScalaCheck version number in `build.sbt` in this repo and running `test:compile`:

```
[error] /Users/tisue/twitter-util/util-stats/src/test/scala/com/twitter/finagle/stats/DelegatingStatsReceiverTest.scala:15:71: value wrap is not a member of object org.scalacheck.Gen
[error]  Note: implicit value impl is not applicable here because it comes after the application point and it lacks an explicit result type
[error]     private[this] val inMemoryStatsReceiver: Gen[StatsReceiver] = Gen.wrap {
[error]                                                                       ^
[error] /Users/tisue/twitter-util/util-stats/src/test/scala/com/twitter/finagle/stats/DelegatingStatsReceiverTest.scala:70:39: value wrap is not a member of object org.scalacheck.Gen
[error]  Note: implicit value impl is not applicable here because it comes after the application point and it lacks an explicit result type
[error]     implicit val impl = Arbitrary(Gen.wrap(statsReceiverTopology(0)))
[error]                                       ^
[error] /Users/tisue/twitter-util/util-stats/src/test/scala/com/twitter/finagle/stats/DelegatingStatsReceiverTest.scala:77:12: could not find implicit value for parameter arbA: org.scalacheck.Arbitrary[com.twitter.finagle.stats.StatsReceiver]
[error]     forAll { statsReceiver: StatsReceiver =>
[error]            ^
```

for now, in the community build, I will disable the compiling & running of your tests

ryanoneill:
Thanks for letting us know @SethTisue. We'll address this as soon as we can. We are using 1.13.4 internally at Twitter. So if we can't write code that can work for both 1.13.4 and 1.14.0, it might be a bit.

SethTisue:
tests re-enabled in community build: scala/community-builds@a0484294680c7b24d04dab68b5a0f41b496f7e64

PR #216, opened 2018-06-07
use Gen.delay instead of Gen.wrap

xuwei-k:
### Problem

`org.scalacheck.Gen.wrap` is deprecated since scalacheck 1.13.0. removed scalacheck 1.14.0

### Solution

use `Gen.delay`

### Result

fix https://github.com/twitter/util/issues/214.
There is no substantial changes because `Gen.wrap` call `Gen.delay`.
https://github.com/rickynils/scalacheck/blob/1.13.4/src/main/scala/org/scalacheck/Gen.scala#L446


yufangong:
merged in https://github.com/twitter/util/commit/91b3a5e7bec60786879f8deba38ee7fc6fcd383d

PR #221, opened 2018-08-06
address deprecations

martijnhoekstra:
Enables and heeds deprecation warnings

Also make some tests platform-independent

Problem

there were deprecation warnings, and tests that failed on Windows

Solution

Remove deprecations

Result

The build will be deprecation-clean, aiding a possible 2.13 migration.

PR #222, opened 2018-08-06
address deprecations

martijnhoekstra:
Enables and heeds deprecation warnings

Also make some tests platform-independent

Enables and heeds deprecation warnings

Also make some tests platform-independent

Problem

there were deprecation warnings, and tests that failed on Windows

Solution

Remove deprecations

Result

The build will be deprecation-clean, aiding a possible 2.13 migration.

Re-issue of https://github.com/twitter/util/pull/221

### DIFF
--- a/util-cache/src/main/scala/com/twitter/cache/Refresh.scala
+++ b/util-cache/src/main/scala/com/twitter/cache/Refresh.scala
@@ -65,6 +65,6 @@ object Refresh {
         }
     }
     // Return result, which is a no-arg function that returns a future
-    result
+    result _
   }
 }
--- a/util-collection/src/main/scala/com/twitter/util/LruMap.scala
+++ b/util-collection/src/main/scala/com/twitter/util/LruMap.scala
@@ -3,7 +3,7 @@ package com.twitter.util
 import java.{util => ju}
 import java.util.LinkedHashMap
 import scala.collection.JavaConverters._
-import scala.collection.mutable.{Map, MapLike, SynchronizedMap}
+import scala.collection.mutable.{Map, MapLike}
 
 /**
  * A wrapper trait for java.util.Map implementations to make them behave as scala Maps.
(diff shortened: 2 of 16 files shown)


util-collection/src/main/scala/com/twitter/util/LruMap.scala
```diff
@@ -64,7 +64,32 @@ class LruMap[K, V](val maxSize: Int, val underlying: ju.Map[K, V])
...
+  override def foreach[U](f: ((K, V)) => U) = synchronized { super.foreach(f) }
+  override def apply(key: K): V = synchronized { super.apply(key) }
+  override def keySet: scala.collection.Set[K] = synchronized { super.keySet }
+  override def keys: scala.collection.Iterable[K] = synchronized { super.keys }
+  override def keysIterator: Iterator[K] = synchronized { super.keysIterator }
+  override def isEmpty: Boolean = synchronized { super.isEmpty }
+  override def contains(key: K): Boolean = synchronized {super.contains(key) }
+  override def isDefinedAt(key: K) = synchronized { super.isDefinedAt(key) }
   override def empty: SynchronizedLruMap[K, V] = new SynchronizedLruMap[K, V](maxSize)
+  //!!! todo: also add all other methods
```
martijnhoekstra: Deprecated trait copy-pasted in, including ominous todo.

util-core/src/main/java/com/twitter/concurrent/Spools.java
```diff
@@ -22,7 +22,7 @@ private Spools() { }
    * Creates a new `Spool` of given `elems`.
    */
   public static <T> Spool<T> newSpool(Collection<T> elems) {
-    Seq<T> seq = JavaConversions.asScalaBuffer(new ArrayList<T>(elems)).toSeq();
+    Seq<T> seq = JavaConverters.collectionAsScalaIterableConverter(elems).asScala().toSeq();
```
mosesn: I don't *love love* this because of the intermediate object, but we do shave the ArrayList alloc, which is pretty good.  can you add a TODO to change to `collectionAsScalaIterable` when we drop support for 2.11.x?

util-core/src/main/java/com/twitter/util/Closables.java
```diff
@@ -44,7 +44,7 @@ private Closables() { }
    */
   public static Closable all(Closable... closables) {
     return Closable$.MODULE$.all(
-      JavaConversions.asScalaBuffer(Arrays.asList(closables))
+      scala.collection.mutable.WrappedArray$.MODULE$.make(closables)
```
mosesn: again, TODO to make this better after we drop 2.11.x?

util-core/src/main/java/com/twitter/util/javainterop/Scala.java
```diff
@@ -32,14 +31,14 @@
     if (jList == null) {
       return scala.collection.immutable.Seq$.MODULE$.<E>empty();
     } else {
-      return asScalaBuffer(jList).toList();
+      return JavaConverters.asScalaBufferConverter(jList).asScala().toList();
```
mosesn: ditto

util-core/src/main/java/com/twitter/util/javainterop/Scala.java
```diff
@@ -51,14 +50,14 @@
     if (jSet == null) {
       return scala.collection.immutable.Set$.MODULE$.<E>empty();
     } else {
-      return asScalaSet(jSet).toSet();
+      return JavaConverters.asScalaSetConverter(jSet).asScala().toSet();
```
mosesn: ditto

util-core/src/test/scala/com/twitter/util/CancellableTest.scala
```diff
@@ -9,7 +9,8 @@ class CancellableTest extends WordSpec {
   "CancellableSink" should {
     "cancel once" in {
       var count = 0
-      val s = new CancellableSink { count += 1 }
+      def increment: Unit = count += 1
+      val s = new CancellableSink(increment)
```
mosesn: ffti (feel free to ignore): `new CancellableSink({ count += 1 })`

util-core/src/main/java/com/twitter/concurrent/Spools.java
martijnhoekstra: Initially, I mainly tried to make this work with the smallest change possible. Thinking about this more, iterating the java collection right into the spool without any converters might be a better strategy.

If I try to do that, I'm going to struggle a bit, since I don't actually know Java, and my knowledge of the JDK is limited to what I regularly use from Scala.

mosesn: yah, take a stab at it.  java is basically a subset of scala.

util-core/src/main/java/com/twitter/util/Closables.java
mosesn: if it compiles, that works for me.

util-collection/src/main/scala/com/twitter/util/LruMap.scala
nepthar: is the TODO still relevant to us? meaning, is it something that _someone_ will do _someday_ here? Perhaps we can just remove it.

util-cache/src/main/scala/com/twitter/cache/Refresh.scala
```diff
@@ -65,6 +65,6 @@ object Refresh {
         }
     }
     // Return result, which is a no-arg function that returns a future
-    result
+    result _
```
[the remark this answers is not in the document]
martijnhoekstra: yes, exactly

martijnhoekstra:
The todo *may* still be relevant to twitter util -- it would take a bit of investigation to figure out whether that's true or not. 

The (unfortunate) case is that that todo has been in the original case for pretty much ever, and the depreciation has to do with the strategy of mixing in "synchronization" traits, which is not a great strategy. Util has always depended on exactly this potentially broken behaviour.

Effectively, the change in the PR is a NOP. Would you like me to take another look at whether I can implement a proper synchronized version of LruMap? If so, within this PR or independent from it?

util-collection/src/main/scala/com/twitter/util/LruMap.scala
martijnhoekstra: The todo may still be relevant to twitter util -- it would take a bit of investigation to figure out whether that's true or not.

The (unfortunate) case is that that todo has been in the original case for pretty much ever, and the depreciation has to do with the strategy of mixing in "synchronization" traits, which is not a great strategy. Util has always depended on exactly this potentially broken behaviour.

Effectively, the change in the PR is a NOP. Would you like me to take another look at whether I can implement a proper synchronized version of LruMap? If so, within this PR or independent from it?

nepthar: Thanks for the background! For this PR, I don't think implementing a proper synchronized version makes sense.

I don't really have a read on whether or not the extra methods would be useful. If this is indeed a noop, we can leave it as is unless someone else feels strongly.

smlance:
Merged in https://github.com/twitter/util/commit/e2aca990a2257b727a2bc3042c0f7651afc1584e. Thanks @martijnhoekstra!

PR #225, opened 2018-09-03
tweak sources to compile on JDK 9

SethTisue:
fixing compilation errors such as:

[error] /Users/tisue/twitter-util/util-core/src/main/scala/com/twitter/io/Buf.scala:873:25: ambiguous reference to overloaded definition,
[error] both method position in class ByteBuffer of type (x$1: Int)java.nio.ByteBuffer
[error] and  method position in class Buffer of type ()Int
[error] match expected type ?
[error]         val limit = dup.position + math.min(until, length)
[error]                         ^

this came up in the Scala 2.12 community build

### DIFF
--- a/util-core/src/main/scala/com/twitter/io/Buf.scala
+++ b/util-core/src/main/scala/com/twitter/io/Buf.scala
@@ -870,9 +870,9 @@ object Buf {
       else if (isSliceIdentity(from, until)) this
       else {
         val dup = underlying.duplicate()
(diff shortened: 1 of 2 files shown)


mosesn:
Love it, thanks @SethTisue! Do you mind updating your description or commit message to use our commit message format?  The scheme looks like

```
Problem

Solution

Result
```

nepthar:
Thanks so much! This merged in a4faaa7

PR #231, opened 2018-09-28
Refactor Duration `*` implementation.

girishjjain:
Problem
Duration multiply implementation code is convoluted. 

Solution
Refactor existing implementation to improve code readability.

### DIFF
--- a/util-core/src/main/scala/com/twitter/util/Duration.scala
+++ b/util-core/src/main/scala/com/twitter/util/Duration.scala
@@ -373,16 +373,23 @@ sealed class Duration private[util] (protected val nanos: Long) extends {
     def overflowedDuration(a: Long, b: Long): Duration =
       if ((a < 0) == (b < 0)) Duration.Top else Duration.Bottom
 
-    // a < b
-    def multiplyNanos(a: Long, b: Long): Duration =
-      if (a < 0L)
-        if (b < 0L && a < Long.MaxValue / b) overflowedDuration(a, b)
-        else if (b > 0L && Long.MinValue / b > a) overflowedDuration(a, b)
-        else Duration.fromNanoseconds(a * b)
-      else if (a > 0L && a > Long.MaxValue / b) overflowedDuration(a, b)
-      else Duration.fromNanoseconds(a * b)
+    def multiplyNanos(a: Long, b: Long): Duration = {
+      val smaller = if (a <= b) a else b
+      val larger = if (a > b) a else b
 
-    if (nanos > x) multiplyNanos(x, nanos) else multiplyNanos(nanos, x)
+      larger match {
+        case l if (l == 0 || smaller == 0) => Duration.fromNanoseconds(0)
+        case l if (l > 0) => smaller match {
+          case s if (s > 0 && Long.MaxValue / l >= smaller) => Duration.fromNanoseconds(a * b)
+          case s if (s < 0 && Long.MinValue / l <= smaller) => Duration.fromNanoseconds(a * b)
+          case _ => overflowedDuration(a, b)
+        }
+        case l if (Long.MaxValue / l) <= smaller => Duration.fromNanoseconds(a * b)
+        case _ => overflowedDuration(a, b)
+      }
+    }
+
+    multiplyNanos(nanos, x)
   }
 
   /**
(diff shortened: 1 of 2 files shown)


ryanoneill:
Hi @girishjjain. Thank you so much for your contribution to Util. It looks like this code specifically was last modified three years ago by @vkostyukov for performance reasons (i.e. a reduction in allocations). 

Before we proceed, can I ask you to run the [DurationBenchmark](https://github.com/twitter/util/blob/develop/util-benchmark/src/main/scala/com/twitter/util/DurationBenchmark.scala) without your change and with it so that we can see the comparison as we're evaluating the change. 

Thanks.

ryanoneill:
My apologies as I meant to link to the actual change from 2015 as well.
https://github.com/twitter/util/commit/4e9a711b9daebbe14c2b0a072cca59552e0dfcf4

girishjjain:
Ok, here're the numbers. First from the base develop branch:
```
Develop branch
[info] Benchmark                                       Mode  Cnt   Score   Error  Units
[info] DurationBenchmark.durationAddDelta              avgt   10   7.162 ± 0.339  ns/op
[info] DurationBenchmark.durationAddDeltaOverflow      avgt   10   5.318 ± 0.110  ns/op
[info] DurationBenchmark.durationEquals                avgt   10   0.632 ± 0.005  ns/op
[info] DurationBenchmark.durationFloor                 avgt   10  13.661 ± 0.366  ns/op
[info] DurationBenchmark.durationMod                   avgt   10  13.792 ± 0.118  ns/op
[info] DurationBenchmark.durationMultiplyLong          avgt   10  13.007 ± 0.273  ns/op
[info] DurationBenchmark.durationMultiplyLongOverflow  avgt   10   3.820 ± 0.315  ns/op
```

and next from my feature branch  
```
girishjjain:refactorDuration
[info] Benchmark                                       Mode  Cnt   Score   Error  Units
[info] DurationBenchmark.durationAddDelta              avgt   10   6.869 ± 0.113  ns/op
[info] DurationBenchmark.durationAddDeltaOverflow      avgt   10   5.307 ± 0.073  ns/op
[info] DurationBenchmark.durationEquals                avgt   10   0.632 ± 0.011  ns/op
[info] DurationBenchmark.durationFloor                 avgt   10  13.856 ± 0.318  ns/op
[info] DurationBenchmark.durationMod                   avgt   10  13.934 ± 0.243  ns/op
[info] DurationBenchmark.durationMultiplyLong          avgt   10  14.609 ± 0.220  ns/op
[info] DurationBenchmark.durationMultiplyLongOverflow  avgt   10   4.010 ± 0.058  ns/op
```

util-core/src/main/scala/com/twitter/util/Duration.scala
```diff
@@ -373,16 +373,23 @@ sealed class Duration private[util] (protected val nanos: Long) extends {
...
-    def multiplyNanos(a: Long, b: Long): Duration =
-      if (a < 0L)
-        if (b < 0L && a < Long.MaxValue / b) overflowedDuration(a, b)
-        else if (b > 0L && Long.MinValue / b > a) overflowedDuration(a, b)
-        else Duration.fromNanoseconds(a * b)
-      else if (a > 0L && a > Long.MaxValue / b) overflowedDuration(a, b)
-      else Duration.fromNanoseconds(a * b)
-
-    if (nanos > x) multiplyNanos(x, nanos) else multiplyNanos(nanos, x)
+    def multiplyNanos(a: Long, b: Long): Duration = {
```
vkostyukov: I'm somewhat skeptical this actually improves readability. Perhaps others might find it more readable.

ryanoneill:
I should have been explicit. Can you run with the 'gc' option as well?
e.g.
```
./sbt 'project util-benchmark' 'jmh:run -prof gc DurationBenchmark'
```
Once we have those numbers, I'll put it up for review internally and if the numbers are the same we'll decide on the readability. Thanks again.

util-core/src/main/scala/com/twitter/util/Duration.scala
girishjjain: I agree that it's not drastic improvement, maybe it reads slightly better? I was also thinking of adding comments at the end of certain lines, as below: 

```
larger match {
  case l if (l == 0 || smaller == 0) => Duration.fromNanoseconds(0)                       // zero
  case l if (l > 0) => smaller match {
    case s if (s > 0 && Long.MaxValue / l >= smaller) => Duration.fromNanoseconds(a * b)  // + times + = + result
    case s if (s < 0 && Long.MinValue / l <= smaller) => Duration.fromNanoseconds(a * b)  // + times - = - result
    case _ => overflowedDuration(a, b)
  }
  case l if (Long.MaxValue / l) <= smaller => Duration.fromNanoseconds(a * b)             // - times - = + result
  case _ => overflowedDuration(a, b)
}
```

ryanoneill:
Hi @girishjjain, after talking it over with a few team members and looking at the performance implications of this patch specifically, we're going to decline to merge this into util. I appreciate the time and effort you put into this especially in regards to providing the benchmarking data that we asked for.

PR #232, opened 2018-10-04
jvm: Use newer jdk management api when the old one is not available

SimonAdameit:
Problem

Newer jdks do not expose `sun.management.ManagementFactory`. The
code to branch between `sun.management.ManagementFactoryHelper` for
jdk7 and `sun.management.ManagementFactory` for earlier jdks only
caught `NoSuchFieldException`. This lead to the `NilJvm` being used
on newer jdks that do not expose `sun.management.ManagementFactory`.

Solution

Catch `Throwable` as is done elsewhere in the same codebase and
switch to the `sun.management.ManagementFactory`

Result

The Hotspot class can be instantiated for newer jdks.

### DIFF
--- a/util-jvm/src/main/scala/com/twitter/jvm/Hotspot.scala
+++ b/util-jvm/src/main/scala/com/twitter/jvm/Hotspot.scala
@@ -9,6 +9,7 @@ import javax.management.openmbean.CompositeDataSupport
 import javax.management.{ObjectName, RuntimeMBeanException}
 import scala.collection.JavaConverters._
 import scala.language.reflectiveCalls
+import scala.util.control.NonFatal
 
 class Hotspot extends Jvm {
   private[this] val epoch =
@@ -32,7 +33,7 @@ class Hotspot extends Jvm {
       // jdk5/6 have jvm field in ManagementFactory class
       Class.forName("sun.management.ManagementFactory").getDeclaredField("jvm")
     } catch {
-      case _: Throwable =>
+      case NonFatal(_) =>
         // jdk7 moves jvm field to ManagementFactoryHelper class
         Class.forName("sun.management.ManagementFactoryHelper").getDeclaredField("jvm")
     }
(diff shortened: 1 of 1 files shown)


ryanoneill:
Hi @SimonAdameit. Thanks for the pull request. I think we've settled on trying not to use 'Throwable' to catch exceptions when possible. Would '[NonFatal](https://www.scala-lang.org/api/2.11.12/index.html#scala.util.control.NonFatal$)' work here instead?

SimonAdameit:
@ryanoneill Sure, I've changed that in the PR. One remaining problem with this code is that since JDK9 access to the `sun.management.ManagementFactoryHelper` is fenced off by the module system. So instead of using the `NilJvm` with the new catch you'll receive a warning: `WARNING: An illegal reflective access operation has occurred` as the package isn't exported by its module. It will still be permitted, but warns that any future JVM may disallow that access. The warning can be disabled via java options, but I'm not sure what tradeoff you want to take here. Also I don't know the proper way to get that information in >=JDK9 without "illegal" access to internals.

SimonAdameit:
So after this change, on JDK10 you'd have to add these options to get rid of the illegal access warnings due to reflecting on non-exported classes/fields:

```scala
javaOptions += "--add-opens=java.management/sun.management=ALL-UNNAMED"
javaOptions += "--add-opens=java.management/sun.management.counter=ALL-UNNAMED"
javaOptions += "--add-opens=java.management/sun.management.counter.perf=ALL-UNNAMED"
```

ryanoneill:
Hi @SimonAdameit, thanks for the update. Let me talk to some of our JVM folks here and see if they have good opinions on how to proceed here.

SimonAdameit:
@ryanoneill I think merging this will improve the situation. It might be worth a note in the documentation that these java options can be used to get rid of the warning when you're using a >= JVM9

bryce-anderson:
This has been merged as of 6f587b35acb19b9e526ff6f0111f9512d091175a. Thanks for the PR @SimonAdameit!

PR #236, opened 2018-11-07
refactor: replace filter.size with more concise count

teresy:
refactors to use the simpler `count` function instead of `filter(...).size`


kevinoliver:
Thanks again @teresy this got merged back in 9f01785ae335d84652473d9174063d47dd912c8e

PR #238, opened 2018-11-22
`Future.???` will return different exception instance each time.

kenchan0130:
## Problem

Currently, `Future.???` returns same instance when execute `Future.???` timing.
Therefore we cannot get valid stacktrace (information on where it is defined).

## Solution

Change `Future.???` define method from val to def.

## Result

This change make us been able to get valid stacktrace.


bryce-anderson:
Merged as sha 3909d1ad17debc65b44a232ef92ca21f924d25e7. Thanks for the PR!

PR #247, opened 2019-04-15
doc: Remove unused parameter in Future conversion example

### LINKED ISSUE #246: Minor - Wrong Code in Cookbook documentation
Wrong Code in Cookbook documentation

### Expected behavior

In page https://twitter.github.io/util/guide/util-cookbook/futures.html , in 1st Code example method  "asScala" has param "implicit e: ExecutionContext " which is not required by code contained in the method "asScala" (Execution Context is not used by contained code as well)

### Actual behavior

method  "asScala" doesn't have  param "implicit e: ExecutionContext "

### Steps to reproduce the behavior

NA

soojee:
**Problem**

`asScala` method in the first example of the Futures documentation passes in an unused parameter `implicit e: ExecutionContext`.

**Solution**

Removed unused parameter. 

Fixes #246.

### DIFF
--- a/doc/src/sphinx/util-cookbook/futures.rst
+++ b/doc/src/sphinx/util-cookbook/futures.rst
@@ -29,7 +29,7 @@ Scala:
 
   /** Convert from a Twitter Future to a Scala Future */
   implicit class RichTwitterFuture[A](val tf: TwitterFuture[A]) extends AnyVal {
-    def asScala(implicit e: ExecutionContext): ScalaFuture[A] = {
+    def asScala(): ScalaFuture[A] = {
       val promise: ScalaPromise[A] = ScalaPromise()
       tf.respond {
         case Return(value) => promise.success(value)


doc/src/sphinx/util-cookbook/futures.rst
```diff
@@ -41,7 +41,7 @@ Scala:
 
   /** Convert from a Scala Future to a Twitter Future */
   implicit class RichScalaFuture[A](val sf: ScalaFuture[A]) extends AnyVal {
-    def asTwitter(implicit e: ExecutionContext): TwitterFuture[A] = {
+    def asTwitter(): TwitterFuture[A] = {
       val promise: TwitterPromise[A] = new TwitterPromise[A]()
       sf.onComplete {
```
yufangong: I think the implicit ExecutionContext is still needed here for `onComplete`.
https://github.com/scala/scala/blob/2.13.x/src/library/scala/concurrent/Future.scala#L121
Do you mind double checking if this is true?

soojee: You're right; `asTwitter` still needs to pass in an implicit ExecutionContext since `onComplete` takes in two parameters: a Try function and an implicit ExecutionContext. 

Thanks for catching this! I'm still new to Scala, and I've learned a lot about implicit parameters tonight! 😃

yufangong:
Thanks, @soojee, we will merge this shortly.

yufangong:
@soojee Thanks again for your contribution! This is merged https://github.com/twitter/util/commit/78a03102f502eb3aa9d1483994f2b6b10193bda9.

PR #248, opened 2019-04-15
Update to 2.13.0

martijnhoekstra:
Aim for (but not always accomplish) source compatibility

Prefer to use "default" `Seq` over `scala.collection.Seq`

Problem

Solution

Align with 2.13.0 collections

Result

Building under 2.13.0

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ before_cache:
 scala:
   - 2.11.12
   - 2.12.8
+  - 2.13.0
 
 jdk:
   - oraclejdk8
--- a/build.sbt
+++ b/build.sbt
@@ -22,7 +22,7 @@ val slf4jApi = "org.slf4j" % "slf4j-api" % slf4jVersion
 
 val defaultProjectSettings = Seq(
   scalaVersion := "2.12.8",
-  crossScalaVersions := Seq("2.11.12", "2.12.8")
+  crossScalaVersions := Seq("2.11.12", "2.12.8", "2.13.0")
 )
 
 val baseSettings = Seq(
--- a/build.sbt
+++ b/build.sbt
@@ -12,12 +12,12 @@ val zkDependency = "org.apache.zookeeper" % "zookeeper" % zkVersion excludeAll(
   ExclusionRule("javax.jms", "jms")
 )
 val slf4jVersion = "1.7.21"
-val jacksonVersion = "2.9.8"
+val jacksonVersion = "2.9.9"
 
 val guavaLib = "com.google.guava" % "guava" % "19.0"
 val caffeineLib = "com.github.ben-manes.caffeine" % "caffeine" % "2.3.4"
 val jsr305Lib = "com.google.code.findbugs" % "jsr305" % "2.0.1"
-val scalacheckLib = "org.scalacheck" %% "scalacheck" % "1.13.4" % "test"
+val scalacheckLib = "org.scalacheck" %% "scalacheck" % "1.14.0" % "test"
 val slf4jApi = "org.slf4j" % "slf4j-api" % slf4jVersion
 
 val defaultProjectSettings = Seq(
@@ -32,10 +32,19 @@ val baseSettings = Seq(
   unmanagedClasspath in Compile += Attributed.blank(new java.io.File("doesnotexist")),
   libraryDependencies ++= Seq(
     // See https://www.scala-sbt.org/0.13/docs/Testing.html#JUnit
+    "org.scala-lang.modules" %% "scala-collection-compat" % "2.1.1",
     "com.novocode" % "junit-interface" % "0.11" % "test",
     "org.mockito" % "mockito-all" % "1.10.19" % "test",
-    "org.scalatest" %% "scalatest" % "3.0.0" % "test"
+    "org.scalatest" %% "scalatest" % "3.0.8" % "test"
   ),
+  fork in Test := true,
+  unmanagedSourceDirectories in Compile += {
+    val sourceDir = (sourceDirectory in Compile).value
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, n)) if n >= 13 => sourceDir / "scala-2.13+"
+      case _ => sourceDir / "scala-2.12-"
+    }
+  },
 
   ScoverageKeys.coverageHighlighting := true,
   resolvers +=
@@ -48,7 +57,8 @@ val baseSettings = Seq(
     "-feature",
     "-encoding", "utf8",
     // Needs -missing-interpolator due to https://issues.scala-lang.org/browse/SI-8761
-    "-Xlint:-missing-interpolator"
+    "-Xlint:-missing-interpolator",
+    "-Yrangepos"
   ),
 
   // Note: Use -Xlint rather than -Xlint:unchecked when TestThriftStructure
@@ -216,7 +226,7 @@ lazy val utilCore = Project(
     caffeineLib % "test",
     scalacheckLib,
     "org.scala-lang" % "scala-reflect" % scalaVersion.value,
-    "org.scala-lang.modules" %% "scala-parser-combinators" % "1.0.4"
+    "org.scala-lang.modules" %% "scala-parser-combinators" % "1.1.2"
   ),
   resourceGenerators in Compile += Def.task {
     val projectName = name.value
@@ -368,7 +378,14 @@ lazy val utilStats = Project(
   sharedSettings
 ).settings(
   name := "util-stats",
-  libraryDependencies ++= Seq(caffeineLib, jsr305Lib, scalacheckLib)
+  libraryDependencies ++= Seq(caffeineLib, jsr305Lib, scalacheckLib) ++ {
+    CrossVersion.partialVersion(scalaVersion.value) match {
+      case Some((2, major)) if major >= 13 =>
+        Seq("org.scala-lang.modules" %% "scala-parallel-collections" % "0.2.0" % "test")
+      case _ =>
+        Seq()
+  }
+}
 ).dependsOn(utilCore, utilLint)
 
 lazy val utilTest = Project(
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,6 +2,5 @@ resolvers += Classpaths.sbtPluginReleases
 resolvers += Resolver.sonatypeRepo("snapshots")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.3.0")
-addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.0-RC13")
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.6.0")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.3.4")
--- a/util-app/src/main/scala/com/twitter/app/App.scala
+++ b/util-app/src/main/scala/com/twitter/app/App.scala
@@ -8,6 +8,7 @@ import java.util.concurrent.atomic.AtomicReference
 import java.util.logging.Logger
 import scala.collection.JavaConverters._
 import scala.collection.mutable
+import scala.collection.Seq
 import scala.util.control.NonFatal
 
 /**
--- a/util-app/src/main/scala/com/twitter/app/App.scala
+++ b/util-app/src/main/scala/com/twitter/app/App.scala
@@ -8,7 +8,6 @@ import java.util.concurrent.atomic.AtomicReference
 import java.util.logging.Logger
 import scala.collection.JavaConverters._
 import scala.collection.mutable
-import scala.collection.Seq
 import scala.util.control.NonFatal
 
 /**
@@ -314,7 +313,7 @@ trait App extends Closable with CloseAwaitably {
     deadline: Time
   ): Future[Unit] = {
     Future
-      .collectToTry(lastExits.asScala.toSeq.map(_.close(deadline)))
+      .collectToTry(lastExits.asScala.map(_.close(deadline)).toSeq)
       .by(shutdownTimer, deadline)
       .transform {
         case Return(results) =>
--- a/util-app/src/main/scala/com/twitter/app/App.scala
+++ b/util-app/src/main/scala/com/twitter/app/App.scala
@@ -313,7 +313,7 @@ trait App extends Closable with CloseAwaitably {
     deadline: Time
   ): Future[Unit] = {
     Future
-      .collectToTry(lastExits.asScala.map(_.close(deadline)).toSeq)
+      .collectToTry(lastExits.asScala.toSeq.map(_.close(deadline)))
       .by(shutdownTimer, deadline)
       .transform {
         case Return(results) =>
--- a/util-app/src/main/scala/com/twitter/app/ClassPath.scala
+++ b/util-app/src/main/scala/com/twitter/app/ClassPath.scala
@@ -161,7 +161,7 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
 
   private def jarClasspath(jarFile: File, manifest: java.util.jar.Manifest): Seq[URI] =
     for {
-      m  <- Option(manifest).toSeq
+      m <- Option(manifest).toSeq
       attr <- Option(m.getMainAttributes.getValue("Class-Path")).toSeq
       el <- attr.split(" ").toSeq
       uri <- uriFromJarClasspath(jarFile, el)
--- a/util-app/src/main/scala/com/twitter/app/ClassPath.scala
+++ b/util-app/src/main/scala/com/twitter/app/ClassPath.scala
@@ -228,7 +228,7 @@ private[app] class LoadServiceClassPath extends ClassPath[ClassPath.LoadServiceI
 
   private[app] def readLines(source: Source): Seq[String] = {
     try {
-      source.getLines().toArray.flatMap { line =>
+      source.getLines().toVector.flatMap { line =>
         val commentIdx = line.indexOf('#')
         val end = if (commentIdx != -1) commentIdx else line.length
         val str = line.substring(0, end).trim
--- a/util-app/src/main/scala/com/twitter/app/ClassPath.scala
+++ b/util-app/src/main/scala/com/twitter/app/ClassPath.scala
@@ -6,6 +6,7 @@ import java.net.{URISyntaxException, URLClassLoader, URI}
 import java.nio.charset.MalformedInputException
 import java.util.jar.{JarEntry, JarFile}
 import scala.collection.mutable
+import scala.collection.Seq
 import scala.collection.JavaConverters._
 import scala.io.Source
 
@@ -94,17 +95,20 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
     buf: mutable.Buffer[CpInfo],
     history: mutable.Set[URI]
   ): Unit = {
-    if (uri.getScheme != "file")
-      return
-
-    val f = new File(uri)
-    if (!(f.exists() && f.canRead))
-      return
-
-    if (f.isDirectory)
-      browseDir(f, loader, "", buf)
-    else
-      browseJar(f, loader, buf, history)
+    if (!history.contains(uri)) {
+      history.add(uri)
+      if (uri.getScheme != "file")
+        return
+  
+      val f = new File(uri)
+      if (!(f.exists() && f.canRead))
+        return
+    
+      if (f.isDirectory)
+        browseDir(f, loader, "", buf)
+      else
+        browseJar(f, loader, buf, history)
+    }
   }
 
   private[this] def browseDir(
@@ -117,8 +121,9 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
       return
 
     for (f <- dir.listFiles)
-      if (f.isDirectory && f.canRead)
+      if (f.isDirectory && f.canRead) {
         browseDir(f, loader, prefix + f.getName + "/", buf)
+      }
       else
         processFile(prefix, f, buf)
   }
@@ -138,10 +143,7 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
 
     try {
       for (uri <- jarClasspath(file, jarFile.getManifest)) {
-        if (!seenUris.contains(uri)) {
-          seenUris += uri
           browseUri0(uri, loader, buf, seenUris)
-        }
       }
 
       for {
@@ -166,11 +168,12 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
 
   private def jarClasspath(jarFile: File, manifest: java.util.jar.Manifest): Seq[URI] =
     for {
-      m <- Option(manifest).toSeq
+      m  <- Option(manifest).toSeq
       attr <- Option(m.getMainAttributes.getValue("Class-Path")).toSeq
-      el <- attr.split(" ")
+      el <- attr.split(" ").toSeq
       uri <- uriFromJarClasspath(jarFile, el)
     } yield uri
+    
 
   private def uriFromJarClasspath(jarFile: File, path: String): Option[URI] =
     try {
--- a/util-app/src/main/scala/com/twitter/app/ClassPath.scala
+++ b/util-app/src/main/scala/com/twitter/app/ClassPath.scala
@@ -6,7 +6,7 @@ import java.net.{URISyntaxException, URLClassLoader, URI}
 import java.nio.charset.MalformedInputException
 import java.util.jar.{JarEntry, JarFile}
 import scala.collection.mutable
-import scala.collection.Seq
+import scala.collection.mutable.Builder
 import scala.collection.JavaConverters._
 import scala.io.Source
 
@@ -52,47 +52,40 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
   protected def ignoredPackages: Set[String]
 
   def browse(loader: ClassLoader): Seq[CpInfo] = {
-    val buf = mutable.Buffer[CpInfo]()
+    val buf = Vector.newBuilder[CpInfo]
     val seenUris = mutable.HashSet[URI]()
 
     for ((uri, loader) <- getEntries(loader)) {
       browseUri0(uri, loader, buf, seenUris)
       seenUris += uri
     }
-    buf
+    buf.result
   }
 
   // package protected for testing
   private[app] def getEntries(loader: ClassLoader): Seq[(URI, ClassLoader)] = {
-    val ents = mutable.Buffer[(URI, ClassLoader)]()
-    val parent = loader.getParent
-    if (parent != null)
-      ents ++= getEntries(parent)
+    val parent = Option(loader.getParent)
+ 
+    val ownURIs: Vector[(URI, ClassLoader)] = for {
+      urlLoader <- Vector(loader).collect {case u: URLClassLoader => u}
+      urls <- Option(urlLoader.getURLs()).toVector
+      url <- urls if url != null
+    } yield (url.toURI -> loader)
 
-    loader match {
-      case urlLoader: URLClassLoader =>
-        Option(urlLoader.getURLs) match {
-          case Some(urls) =>
-            urls.foreach { url =>
-              if (url != null)
-                ents += (url.toURI -> loader)
-            }
-          case _ =>
-        }
-      case _ =>
-    }
+    val p = parent.toSeq.flatMap(getEntries)
 
-    ents
+    ownURIs ++ p
+
   }
 
   // package protected for testing
-  private[app] def browseUri(uri: URI, loader: ClassLoader, buf: mutable.Buffer[CpInfo]): Unit =
+  private[app] def browseUri(uri: URI, loader: ClassLoader, buf: Builder[CpInfo, Seq[CpInfo]]): Unit =
     browseUri0(uri, loader, buf, mutable.Set[URI]())
 
   private[this] def browseUri0(
     uri: URI,
     loader: ClassLoader,
-    buf: mutable.Buffer[CpInfo],
+    buf: Builder[CpInfo, Seq[CpInfo]],
     history: mutable.Set[URI]
   ): Unit = {
     if (!history.contains(uri)) {
@@ -115,7 +108,7 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
     dir: File,
     loader: ClassLoader,
     prefix: String,
-    buf: mutable.Buffer[CpInfo]
+    buf: Builder[CpInfo, Seq[CpInfo]]
   ): Unit = {
     if (ignoredPackages.contains(prefix))
       return
@@ -128,12 +121,12 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
         processFile(prefix, f, buf)
   }
 
-  protected def processFile(prefix: String, file: File, buf: mutable.Buffer[CpInfo]): Unit
+  protected def processFile(prefix: String, file: File, buf: Builder[CpInfo, Seq[CpInfo]]): Unit
 
   private def browseJar(
     file: File,
     loader: ClassLoader,
-    buf: mutable.Buffer[CpInfo],
+    buf: Builder[CpInfo, Seq[CpInfo]],
     seenUris: mutable.Set[URI]
   ): Unit = {
     val jarFile = try new JarFile(file)
@@ -163,7 +156,7 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
   protected def processJarEntry(
     jarFile: JarFile,
     entry: JarEntry,
-    buf: mutable.Buffer[CpInfo]
+    buf: Builder[CpInfo, Seq[CpInfo]]
   ): Unit
 
   private def jarClasspath(jarFile: File, manifest: java.util.jar.Manifest): Seq[URI] =
@@ -199,7 +192,7 @@ private[app] class FlagClassPath extends ClassPath[ClassPath.FlagInfo] {
   protected def processFile(
     prefix: String,
     file: File,
-    buf: mutable.Buffer[ClassPath.FlagInfo]
+    buf: Builder[ClassPath.FlagInfo, Seq[ClassPath.FlagInfo]]
   ): Unit = {
     val name = file.getName
     if (isClass(name)) {
@@ -210,7 +203,7 @@ private[app] class FlagClassPath extends ClassPath[ClassPath.FlagInfo] {
   protected def processJarEntry(
     jarFile: JarFile,
     entry: JarEntry,
-    buf: mutable.Buffer[ClassPath.FlagInfo]
+    buf: Builder[ClassPath.FlagInfo, Seq[ClassPath.FlagInfo]]
   ): Unit = {
     val name = entry.getName
     if (isClass(name)) {
@@ -251,7 +244,7 @@ private[app] class LoadServiceClassPath extends ClassPath[ClassPath.LoadServiceI
   protected def processFile(
     prefix: String,
     file: File,
-    buf: mutable.Buffer[ClassPath.LoadServiceInfo]
+    buf: Builder[ClassPath.LoadServiceInfo, Seq[ClassPath.LoadServiceInfo]]
   ): Unit = {
     for (iface <- ifaceOfName(prefix + file.getName)) {
       val source = Source.fromFile(file, "UTF-8")
@@ -263,7 +256,7 @@ private[app] class LoadServiceClassPath extends ClassPath[ClassPath.LoadServiceI
   protected def processJarEntry(
     jarFile: JarFile,
     entry: JarEntry,
-    buf: mutable.Buffer[ClassPath.LoadServiceInfo]
+    buf: Builder[ClassPath.LoadServiceInfo, Seq[ClassPath.LoadServiceInfo]]
   ): Unit = {
     for (iface <- ifaceOfName(entry.getName)) {
       val source = Source.fromInputStream(jarFile.getInputStream(entry), "UTF-8")
--- a/util-core/src/main/scala-2.13+/twitter/util/HealthyQueue.scala
+++ b/util-core/src/main/scala-2.13+/twitter/util/HealthyQueue.scala
@@ -1,0 +1,42 @@
+package com.twitter.util
+
+import scala.collection.mutable
+
+private[util] class HealthyQueue[A](makeItem: () => Future[A], numItems: Int, isHealthy: A => Boolean)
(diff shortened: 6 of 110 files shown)


util-core/src/main/scala/com/twitter/util/BatchExecutor.scala
```diff
@@ -30,7 +31,7 @@ private[util] class BatchExecutor[In, Out](
   sizeThreshold: Int,
   timeThreshold: Duration = Duration.Top,
   sizePercentile: => Float = 1.0f,
-  f: Seq[In] => Future[Seq[Out]]
+  f: Iterable[In] => Future[Seq[Out]]
```
martijnhoekstra: This raises, rather than lowers the requirements. Still, this seems reasonable in the name of unbenchmarked performance.

util-core/src/test/scala/com/twitter/concurrent/AsyncStreamTest.scala
```diff
@@ -389,12 +389,14 @@ class AsyncStreamTest extends FunSuite with GeneratorDrivenPropertyChecks {
       }
     }
 
-    test(s"$impl: buffer() works like Seq.splitAt") {
-      forAll { (items: List[Char], bufferSize: Int) =>
-        val (expectedBuffer, expectedRest) = items.splitAt(bufferSize)
-        val (buffer, rest) = await(fromSeq(items).buffer(bufferSize))
-        assert(expectedBuffer == buffer)
-        assert(expectedRest == toSeq(rest()))
+    test(s"$impl: buffer() works like Seq.splitAt for positive values") {
```
martijnhoekstra: Not sure what to do for negative values, see also https://github.com/scala/bug/issues/11482

martijnhoekstra:
More errors than I thought on different cross-versions. Closing for now, until there is something closer.

martijnhoekstra:
reopening to see if CI agrees more with local now

util-core/src/main/scala/com/twitter/util/Event.scala
```diff
@@ -304,20 +304,27 @@ trait Event[+T] { self =>
...
       }
     }
   }
 
+  /**
+   * Progressively build a Seq of events using the passed-in
+   * builder. A value containing the current version of the collection
+   * is notified for each incoming event.
+   */
+  def build[U >: T, That <: Seq[U]](implicit factory: Factory[U, That]): Event[That] = buildAny(factory)
```
martijnhoekstra: add a seq-specific one, so that build that picks up any ambient factory/cbf builds at least a Seq. If we don't do this, on 2.13.0-RC1, we'll pick up a `Factory[T, Array[T]]` from the companion object of Factory, and build an `Event[Array[T]]` which is not a subtype of `Event[Seq[T]]`

util-core/src/main/scala/com/twitter/util/Event.scala
```diff
@@ -304,20 +304,27 @@ trait Event[+T] { self =>
...
-   * Progressively build a collection of events using the passed-in
+   * Progressively build any collection of events using the passed-in
    * builder. A value containing the current version of the collection
    * is notified for each incoming event.
    */
-  def build[U >: T, That](implicit cbf: CanBuild[U, That]): Event[That] = new Event[That] {
+  def buildAny[That](implicit factory: Factory[T, That]): Event[That] = new Event[That] {
     def register(s: Witness[That]): Closable = {
-      val b = cbf()
+      val b = factory.newBuilder
```
martijnhoekstra: The contract of builder makes calling `result()` more than once undefined behaviour (with the exception of `ReusableBuilder` which can call `result()` again after calling `clear`). That means that this relies upon (and always has relied upon) builder being well-behaved in cases where it doesn't have to play that nice according to its contract.

build.sbt
```diff
@@ -17,25 +17,35 @@ val jacksonVersion = "2.9.6"
...
-  crossScalaVersions := Seq("2.11.12", "2.12.8")
+  crossScalaVersions := Seq("2.11.12", "2.12.8", "2.13.0-RC1")
 )
 
 val baseSettings = Seq(
   version := releaseVersion,
   organization := "com.twitter",
   // Workaround for a scaladoc bug which causes it to choke on empty classpaths.
   unmanagedClasspath in Compile += Attributed.blank(new java.io.File("doesnotexist")),
+  scalafixDependencies += "org.scala-lang.modules" %% "scala-collection-migrations" % "1.0.0",
```
martijnhoekstra: WIP noise, to be removed

build.sbt
```diff
@@ -432,3 +450,5 @@ lazy val utilZkTest = Project(
   name := "util-zk-test",
   libraryDependencies += zkDependency
 ).dependsOn(utilCore % "test")
+
+scalafixDependencies in ThisBuild += "org.scala-lang.modules" %% "scala-collection-migrations" % "1.0.0"
```
martijnhoekstra: WIP noise, to be removed

project/plugins.sbt
```diff
@@ -5,3 +5,5 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.3.0")
 addSbtPlugin("io.get-coursier" % "sbt-coursier" % "1.0.0-RC13")
 addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
 addSbtPlugin("pl.project13.scala" % "sbt-jmh" % "0.3.4")
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.4")
```
martijnhoekstra: WIP noise, to be removed

util-app/src/main/scala/com/twitter/app/ClassPath.scala
```diff
@@ -51,84 +52,81 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
...
-              if (url != null)
-                ents += (url.toURI -> loader)
-            }
-          case _ =>
-        }
-      case _ =>
-    }
+    val parent = Option(loader.getParent)
+ 
+    val ownURIs: Vector[(URI, ClassLoader)] = for {
```
martijnhoekstra: simplified to get rid of the mutable buffer

util-app/src/main/scala/com/twitter/app/ClassPath.scala
```diff
@@ -51,84 +52,81 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
...
+    val p = parent.toSeq.flatMap(getEntries)
+
+    ownURIs ++ p
 
-    ents
   }
 
   // package protected for testing
-  private[app] def browseUri(uri: URI, loader: ClassLoader, buf: mutable.Buffer[CpInfo]): Unit =
+  private[app] def browseUri(uri: URI, loader: ClassLoader, buf: Builder[CpInfo, Seq[CpInfo]]): Unit =
```
martijnhoekstra: general pattern: use a `Builder` and return the immutable result rather than mutate a buffer

util-app/src/main/scala/com/twitter/app/ClassPath.scala
```diff
@@ -51,84 +52,81 @@ private[app] sealed abstract class ClassPath[CpInfo <: ClassPath.Info] {
...
-
-    val f = new File(uri)
-    if (!(f.exists() && f.canRead))
-      return
-
-    if (f.isDirectory)
-      browseDir(f, loader, "", buf)
-    else
-      browseJar(f, loader, buf, history)
+    if (!history.contains(uri)) {
```
martijnhoekstra: do URI checking local to this function -- forking in test resulted in duplicate CpInfo's in the result

util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
```diff
@@ -720,6 +717,27 @@ object AsyncStream {
     }
   }
 
+  private def mergeSeq[A](s: Seq[AsyncStream[A]]): AsyncStream[A] = {
```
martijnhoekstra: Act of desperation -- couldn't get varargs to work like it should

util-core/src/main/scala/com/twitter/io/Reader.scala
```diff
@@ -157,8 +157,8 @@ object Reader {
     require(chunkSize > 0, s"chunkSize should be > 0 but was $chunkSize")
 
     @tailrec
-    private def loop(acc: Seq[Buf], in: Buf): Seq[Buf] = {
-      if (in.length < chunkSize) acc :+ in
+    private def loop(acc: ListBuffer[Buf], in: Buf): Seq[Buf] = {
```
martijnhoekstra: this isn't getting any prettier. I'd be happy to replace with a non-mutable loop, but went for minimal invasiveness

util-core/src/test/java/com/twitter/concurrent/SpoolCompilationTest.java
```diff
@@ -10,6 +10,7 @@
 import java.util.Collection;
 
 public class SpoolCompilationTest {
+
```
martijnhoekstra: whitespace junk, to be reverted

util-core/src/test/scala/com/twitter/io/PipeTest.scala
```diff
@@ -295,23 +300,26 @@ class PipeTest extends FunSuite with Matchers {
     assert(!cf.isDone)
     assertReadEofAndClosed(rw)
   }
+  test("close not satisfied until reads are fulfilled")(closeNotSatisfiedUntillAllReadsDone)
 
-  test("close while read pending") {
+  def closeWhileReadPending = {
```
martijnhoekstra: Some of these had to be factored out, for reasons that aren't clear to me, and may be bugs in assert or scalac: it seems to have to do with macros re-evaluating/re-capturing variables that trip up scalac.

util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
```diff
@@ -208,14 +208,10 @@ class InMemoryStatsReceiver extends StatsReceiver with WithHistogramDetails {
...
-            nearestPosInt(x)
-          }
-          .groupBy(identity)
-          .mapValues(_.size)
+          .groupBy(nearestPosInt)
+          .map { case (k, vs) => BucketAndCount(k, k+1, vs.size) }
           .toSeq
-          .sortWith(_._1 < _._1)
-          .map { case (k, v) => BucketAndCount(k, k + 1, v) }
+          .sortBy(_.lowerLimit)
```
martijnhoekstra: simplified after eliminating the deprecated mapValues

util-stats/src/test/java/com/twitter/finagle/stats/StatsReceiverCompilationTest.java
```diff
@@ -95,18 +100,18 @@ public boolean isNull() {
...
-      public Stat stat(Verbosity verbosity, Seq<String> name) {
-        return nullSr.stat(name);
+      public Counter counter(Verbosity verbosity, scala.collection.Seq<String> name){
+        return nullSr.counter(verbosity, name.toVector());
       }
 
       @Override
-      public Gauge addGauge(Verbosity verbosity, Seq<String> name, Function0<Object> f) {
-        return nullSr.addGauge(verbosity, name, f);
+      public Gauge addGauge(Verbosity verbosity, scala.collection.Seq<String> name, Function0<Object> f) {
```
martijnhoekstra: I don't know how to fix this. I haven't found a way to have multiple parameter list with a varargs that is source-compatible for overriding between 2.13+ and 2.12-. If anyone has ideas, I'm happy to implement them, but for now, I'm out of ideas.

util-core/src/test/scala/com/twitter/io/PipeTest.scala
martijnhoekstra: todo: check if this is still needed in RC3

util-stats/src/test/java/com/twitter/finagle/stats/StatsReceiverCompilationTest.java
martijnhoekstra: implementing both without an override annotation works. This pattern can be further expanded: an abstract go between with both, one final forwarding to the abstract other.

Which one is better to finalize/forward is debatable.

util-core/src/test/scala/com/twitter/io/PipeTest.scala
martijnhoekstra: reported at https://github.com/scala/bug/issues/11556

util-stats/src/main/scala/com/twitter/finagle/stats/StatsReceiver.scala
```diff
@@ -253,4 +256,22 @@ trait StatsReceiver {
   }
 }
 
-abstract class AbstractStatsReceiver extends StatsReceiver
+abstract class AbstractStatsReceiver extends StatsReceiver {
```
martijnhoekstra: In order to make overriding varargs from Java source compatible between 2.12.x and 2.13.x, made the varargs methods final, and made a non-varargs non-aliased abstract method with a `scala.collection.Seq` argument as an extension point for overriding.

.travis.yml
```diff
@@ -26,6 +26,7 @@ before_cache:
 scala:
   - 2.11.12
   - 2.12.8
+  - 2.13.0-RC3
```
intracer: It's time to target 2.13.0

martijnhoekstra: I wish it were, but I'm still blocked on (at least) https://github.com/FasterXML/jackson-module-scala/pull/417

martijnhoekstra:
rebased to resolve conflicts on SeqReader, that now that I come to think of it maybe should be renamed in this PR.

martijnhoekstra:
Semi-random ping, @kevinoliver what can I do to get some review on this? If I want to port other twitter projects to 2.13 this is the bottom rung that all else depends on.

kevinoliver:
Hi @martijnhoekstra 

Thanks. I'll defer to @yufangong and @mosesn who have some interest in this area. Personally, I'd love to see us drop support for 2.11 and move to 2.12 + 2.13.

martijnhoekstra:
Thanks! I didn't have to do anything for 2.11 support specifically in this
PR, but Ican imagine there being good other reasons for that.

util-core/src/main/scala/com/twitter/util/Activity.scala
```diff
@@ -2,12 +2,13 @@ package com.twitter.util
...
 import java.util.{List => JList}
 
-import scala.collection.generic.CanBuild
 import scala.collection.JavaConverters._
 import scala.collection.mutable.Buffer
 import scala.language.higherKinds
 import scala.reflect.ClassTag
 import scala.util.control.NonFatal
+import scala.Iterable
+import scala.collection.compat._
```
yufangong: Which change requires us to enable this? What process can deduce that we need to have this import?
We are thinking about having a scalafix rule to add this import, wonder if it is doable.

util-cache-guava/src/test/scala/com/twitter/cache/guava/LoadingFutureCacheTest.scala
```diff
@@ -4,7 +4,7 @@ import com.google.common.cache.{CacheLoader, CacheBuilder}
 import com.twitter.cache.AbstractLoadingFutureCacheTest
 import com.twitter.util.Future
 import org.junit.runner.RunWith
-import org.scalatest.junit.JUnitRunner
+import org.scalatestplus.junit.JUnitRunner
```
yufangong: We can actually remove this along with 
`org.junit.runner.RunWith`
and 
`@RunWith(classOf[JUnitRunner])`
but we will still need to update other things in this package like
`org.scalatestplus.mockito.MockitoSugar`

util-core/src/main/scala-2.13+/twitter/util/HealthyQueue.scala
```diff
@@ -0,0 +1,42 @@
+package com.twitter.util
```
yufangong: we will need to find out if this also works for pants build.

util-core/src/main/scala/com/twitter/util/Activity.scala
yufangong: Also, will that be any compile error without this import and build against 2.12?

yufangong:
Thank you @martijnhoekstra so much! I will continue reviewing this.
We are using two build systems for util, internally [pants](https://www.pantsbuild.org/1.0.html) and externally sbt. I would like to know how scala-collection-compat works?
I assume we have to enable it to build against 2.12? I'll imagine we will have pants to work this scala-collection-compat works and build against 2.12.X, and sbt cross-build 2.12 + 2.13.

cc @mosesn and @ShaneDelmore

util-core/src/main/scala-2.13+/twitter/util/HealthyQueue.scala
martijnhoekstra: I'll see if I can get a Pants setup working, and read up about how it selects which files to build for which targets.

 Would you happen to know whether it works on WSL?

util-core/src/main/scala/com/twitter/util/Activity.scala
martijnhoekstra: @yufangong 2.12 doesn't have `Factory`, its added by compat

util-cache-guava/src/test/scala/com/twitter/cache/guava/LoadingFutureCacheTest.scala
martijnhoekstra: Do you want to remove `@RunWith(classOf[JUnitRunner])` globally?

martijnhoekstra:
@yufangong scala compat mostly adds extension methods and type aliases to make it possible to work with scala collections in a source compatible manner, particularly around `CanBuildFrom`, and offers `ArraySeq` that provides an immutable facade over `Array` (and may be used for example to unsafely wrap a varargs array, so it can passed as an immutable Seq which is needed for 2.13 varargs  under the assumption it's not going to be changed) . It's a relatively small library, and can be found here: https://github.com/scala/scala-collection-compat

martijnhoekstra:
@yufangong also, what's the correct way to run the benchmarks? I've tried to be performance-sensitive where it matters, but changes are big enough for something to may have regressed -- especially with `toSeq` (or who knows, progressed).

util-core/src/test/scala/com/twitter/io/PipeTest.scala
yufangong: I ran into issues with PipeTest also in my previous minor experiment, subscribed to the bug.

util-core/src/main/scala-2.13+/twitter/util/HealthyQueue.scala
yufangong: Pants confirmed we can do similar things as sbt to point to the 213 directories, I think we can do it internally when merging this.

util-cache-guava/src/test/scala/com/twitter/cache/guava/LoadingFutureCacheTest.scala
yufangong: We do, but should do it in another effort. I think we can remove these either before this change or after (assuming less conflicts for you)?

yufangong:
@martijnhoekstra, all benchmarks sit in util-benchmark/.
here's one example running them https://github.com/twitter/util/blob/develop/util-benchmark/src/main/scala/com/twitter/io/BufBenchmark.scala#L9

util-core/src/test/scala/com/twitter/io/PipeTest.scala
martijnhoekstra: Factoring the test body into a method consistently worked around for now, if you run into it elsewhere.

util-core/src/main/scala-2.13+/twitter/util/HealthyQueue.scala
martijnhoekstra: I just got the opposite answer from the pants slack from stu hood, who suggested an issue against pants, and different branches as a workaround, so that's a relief

martijnhoekstra:
I'll run them later tonight, when my laptop is otherwise idle, in the
configurations 2.11 and 2.12 pre PR for a baseline, and then 2.11, 2.13 and
2.13 on this branch to compare. Fingers crossed.

util-cache-guava/src/test/scala/com/twitter/cache/guava/LoadingFutureCacheTest.scala
martijnhoekstra: If we remove before this PR, it will significantly reduce the number of files touched in this PR and with that its signal to noise ratio. The rebase against that will be big but easy for this PR. Happy to do that. WDYT?

yufangong: I think that sounds perfect!

util-core/src/main/scala-2.13+/twitter/util/HealthyQueue.scala
yufangong: Yes, that won't block this is what I heart. so 👍 

>Would you happen to know whether it works on WSL?
What is WSL?

martijnhoekstra: WSL is the Linux that comes with Windows, so I can run Pants locally without spinning up a VM, but if I can make the whole pants thing someone else's problem for the extent of this PR, then I'm all good.

util-cache-guava/src/test/scala/com/twitter/cache/guava/LoadingFutureCacheTest.scala
martijnhoekstra: PR for that is here: https://github.com/twitter/util/pull/249

If CI is green there (and actually runs all tests), and can be merged, we can rebase this and reduce the diff size a notch or two.

martijnhoekstra:
That benchmark run will be complete in 10.5 days.

I can't spare hardware to chug on this for that amount of time. I'll try to isolate the likeliest things to have regressed, and run limited runs on those, but that's going to be very limited.

Can you see if hardware can be allocated from your side to chug through the bench suite for all benchmarks for all 5 configurations?

martijnhoekstra:
Based on bench results, I optimized `Future#CollectToTry` for cases with 0 or 1 elements.

I have no idea what's going on with the `PollThenOffer` benchmark on `AsyncStream`. 

With an invocation with many, long runs, I have massive uncertainty for all configurations. For this branch, on 2.13.0 the results are

```
sbt:util-benchmark> jmh:run .*alternatePoll.* -i 15 -r 15 -w 15 -wi 15 -rf json -rff poll213.json
[info] Running (fork) org.openjdk.jmh.Main .*alternatePoll.* -i 15 -r 15 -w 15 -wi 15 -rf json -rff [
    {
        "jmhVersion" : "1.21",
        "benchmark" : "com.twitter.concurrent.AsyncQueueBenchmark.alternatePollThenOffer",
        "mode" : "avgt",
        "threads" : 10,
        "forks" : 1,
        "jvm" : "C:\\Program Files\\Java\\jdk1.8.0_201\\jre\\bin\\java.exe",
        "jvmArgs" : [
        ],
        "jdkVersion" : "1.8.0_201",
        "vmName" : "Java HotSpot(TM) 64-Bit Server VM",
        "vmVersion" : "25.201-b09",
        "warmupIterations" : 15,
        "warmupTime" : "15 s",
        "warmupBatchSize" : 1,
        "measurementIterations" : 15,
        "measurementTime" : "15 s",
        "measurementBatchSize" : 1,
        "primaryMetric" : {
            "score" : 124036.1660659384,
            "scoreError" : 44229.925442361506,
            "scoreConfidence" : [
                79806.2406235769,
                168266.09150829993
            ],
            "scorePercentiles" : {
                "0.0" : 71873.69900185612,
                "50.0" : 107590.91232334396,
                "90.0" : 181831.49408249106,
                "95.0" : 184684.88575997873,
                "99.0" : 184684.88575997873,
                "99.9" : 184684.88575997873,
                "99.99" : 184684.88575997873,
                "99.999" : 184684.88575997873,
                "99.9999" : 184684.88575997873,
                "100.0" : 184684.88575997873
            },
            "scoreUnit" : "ns/op",
            "rawData" : [
                [
                    71873.69900185612,
                    79431.00469862472,
                    94765.7680139789,
                    79963.13558617413,
                    96172.37546569874,
                    91656.0104461496,
                    104439.12600294172,
                    108561.26675847107,
                    107590.91232334396,
                    165730.96178169094,
                    165869.08503153728,
                    171960.7776593581,
                    179929.23296416595,
                    184684.88575997873,
                    157914.24949510623
                ]
            ]
        },
        "secondaryMetrics" : {
        }
    }
]
```

I have no idea how to explain this massive drop in performance after the first few runs. This is not unique to this port though, the same behaviour shows in the `develop` branch.

martijnhoekstra:
Benchmark results in https://gist.github.com/martijnhoekstra/3303264b113a430cf762b7d3d3ab0d4c

Baseline are the 212_1_pre_* files, which is the 2.12 build against the status quo before this PR.
213_* and 212_2_post_* files are the same benchmarks after this PR.

The ones I ran seem to be the most important benches, but I had to be a bit selective.

I didn't run the benchmarks for 2.11

As of this post, they are already a bit out of date -- I prevented some array copying (in general) and unrolled the loop for Future collectToTry to avoid allocation of builder and iterator when possible.

I still want to investigate the following results (2.12 pre PR -> 2.13 post PR)

[] AsyncStreamBenchmark - map - cons -10  -15% (high spread)
[] AsyncStreamBenchmark - $plus$plus - embed - 10 -13% (high spread)
[] AsyncStreamBenchmark - mapConcurrent - embed - 10  -8% (high spread)
[] CumulativeGaugeBenchmark - addGauge  -14% | num=1000

util-cache-guava/src/test/scala/com/twitter/cache/guava/LoadingFutureCacheTest.scala
martijnhoekstra: I've rebased on top of that one. The commit history of this PR is rather messy, and I'd like to rebase and squash into fewer commits to simplify review. If you want, we can split this PR in a couple of different ones along the same way we did this one:

* Update build file and packages without adding a 2.13 build
* Fixes of `Unit` used accidentally as `()`
* Update scalatestplus deprecated aliases
* Add the 2.13 build, and all other changes
* Potentially, fix the deluge of collection converter deprecation warnings when collectionCompat 1.2.0 is released (any day/hour now)

yufangong:
> That benchmark run will be complete in 10.5 days.
> 
> I can't spare hardware to chug on this for that amount of time. I'll try to isolate the likeliest things to have regressed, and run limited runs on those, but that's going to be very limited.
> 
> Can you see if hardware can be allocated from your side to chug through the bench suite for all benchmarks for all 5 configurations?

Yes, I am pulling in this branch and start working on the internal build as well as benchmarks.

martijnhoekstra:
Thats really good news. I'll quit any further crazy rebases then, at least
until you have more to share on that front.

yufangong:
Updates:
We are able to make pants work with this branch yay! Currently, there are some upstream code failures against this, but it is not very bad, and we are working on resolving those. 

We will run a few more benchmarks in addition to the ones you have run. Thanks a ton for doing this!
Do you have any updates to the AsyncStreamBenchmark and CumulativeGaugeBenchmark?

yufangong:
Interpreted from https://gist.github.com/martijnhoekstra/3303264b113a430cf762b7d3d3ab0d4c:
### util-benchmarks:

concurrent

- AsyncQueueBenchmark - done
           alternatePollThenOffer (may because this benchmark is not reliable)
- AsyncSemaphoreBenchmark - done - green results
- AsyncStreamBenchmark - done 
- BrokerBenchmark
- ConduitSpscBenchmark
- OfferBenchmark
- OnceBenchmark
- SchedulerBenchmark

finagle

- CumulativeGaugeBenchmark - done
- StatBenchmark

hashing

- HashableBench
- KetamaDistributorBench
- KeyHasherBenchmark

io

- BufBenchmark
- BufConcatBenchmark
- PipeBenchmark

jvm

- CpuProfileBenchmark

string

- StringConcatenationBenchmark

util

- DurationBenchmark
- FutureBenchmark (CollectToTry optimized) - done - green result
- LocalBenchmark
- MonitorBenchmark
- PromiseBenchmark
- StdBenchAnnotations
- StopwatchBenchmark
- TimeBenchmark
- TimerBenchmark
- TryBenchmark

martijnhoekstra:
The Cumulative Gauge benchmarks were looking good now. The 3 AsyncStream ones had too much jitter in the run I did to be of value. Can you include them on your run?

Note that the benchmarks I ran in many cases didn't have the runtime nor the number of warmup and measurement iterations recommended in the tests, simply for the sake of not taking days. I basically cut every corner that could be cut without crashing or burning. I would still have expected regressions to at least show up, but possibly evaporate under more scrutiny, but more principled runs for everything are welcome.

util-core/src/main/scala/com/twitter/util/Future.scala
```diff
@@ -1404,7 +1422,7 @@ def join[%s](%s): Future[(%s)] = join(Seq(%s)).map { _ => (%s) }""".format(
     sizeThreshold: Int,
     timeThreshold: Duration = Duration.Top,
     sizePercentile: => Float = 1.0f
-  )(f: Seq[In] => Future[Seq[Out]]
+  )(f: Iterable[In] => Future[Seq[Out]]
```
yufangong: This is a breaking change, just want to make sure this is necessary. If so we should add an entry to CHANGELOG(i can add it on my branch).

martijnhoekstra: Maybe it's better to keep this at scala.collection.Seq then.

* For 2.13 it *may* be a breaking change, but some refactoring is already expected there.
* For 2.12.x it won't break anything.

I will investigate, and come back to it this Tuesday, as I'm heading into a long weekend.

martijnhoekstra: https://github.com/twitter/util/pull/248/commits/b5b124bddac097e1044e68a29c266964ac5a16b1 is a way to do it without the Iterable

martijnhoekstra: The tests seem to be mocking me over that change.

I've never used any mocking frameworks. What's the correct way to fix this? Include function composition in the mock?

yufangong: Hey @martijnhoekstra, I modified tests to mock the composed method, and verify against `m`. 
```
val m = mock[Any => Future[Seq[Int]]]
val f = new Predef.Function[Seq[Int], Future[Seq[Int]]] {
  override def apply(v1: Seq[Int]): Future[Seq[Int]] = ???
  override def compose[A](g: A => Seq[Int]): A => Future[Seq[Int]] = m
}
```

I have your branch internally almost ready, passed all tests but only running the rest of benchmarks.

util-core/src/main/scala/com/twitter/util/StorageUnit.scala
```diff
@@ -126,7 +126,7 @@ class StorageUnit(val bytes: Long) extends Ordered[StorageUnit] {
   def max(other: StorageUnit): StorageUnit =
     if (this > other) this else other
 
-  override def toString(): String = inBytes + ".bytes"
+  override def toString(): String = s"$inBytes bytes"
```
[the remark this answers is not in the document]
martijnhoekstra: for the format, or for the interpolator? I missed the `.` there, that change was unintentional, but the interpolator is there because any2stringadd is on its way out, and should be either `s"$inBytes.bytes"` with, or `inBytes.toString + ".bytes"` without interpolator

util-stats/src/main/scala/com/twitter/finagle/stats/StatsReceiver.scala
```diff
@@ -253,4 +256,19 @@ trait StatsReceiver {
...
+  @varargs
+  final override def counter(verbosity: Verbosity, name: String*): Counter = counterImpl(verbosity, name)
+  protected def counterImpl(verbosity: Verbosity, name: scala.collection.Seq[String]): Counter
+  @varargs
+  final override def stat(name: String*): Stat = stat(Verbosity.Default, name: _*)
+  @varargs
+  final override def stat(verbosity: Verbosity, name: String*): Stat = statImpl(verbosity, name)
+  protected def statImpl(verbosity: Verbosity, name: scala.collection.Seq[String]): Stat
+  
+  final override def addGauge(name: String*)(f: => Float): Gauge = addGauge(Verbosity.Default, name: _*)(f)
```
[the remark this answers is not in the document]
martijnhoekstra: @varargs makes little sense if the last argument list isn't the varargs list. Shall I check if they actually generate any different bytecode?

build.sbt
```diff
@@ -12,17 +12,17 @@ val zkDependency = "org.apache.zookeeper" % "zookeeper" % zkVersion excludeAll(
   ExclusionRule("javax.jms", "jms")
 )
 val slf4jVersion = "1.7.21"
-val jacksonVersion = "2.9.8"
+val jacksonVersion = "2.9.9"
```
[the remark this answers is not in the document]
martijnhoekstra: There is no `2.9.8` for 2.13.0, so we have to bump it independently, we have to bump it before and rebase this on top of that.

I have a branch where that's done --https://github.com/twitter/util/compare/develop...martijnhoekstra:pre213 is a diff of a reasonable bump of dependencies and elimination of deprecations that go along with that. But to be honest, I'm not sure what value that brings, other than better baseline for benchmarking maybe? I doubt it would matter much.

util-stats/src/main/scala/com/twitter/finagle/stats/StatsReceiver.scala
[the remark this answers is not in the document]
martijnhoekstra: It does produce different bytecode.

The code

```
object Test {
  @varargs def withAnnotation(args: Int*)(extra: String) = args.length
}

```
produces extra an extra forwarder method:

```
  public int withAnnotation(int..., java.lang.String);
    descriptor: ([ILjava/lang/String;)I
    flags: ACC_PUBLIC, ACC_VARARGS
    Code:
      stack=3, locals=3, args_size=3
         0: aload_0
         1: getstatic     #31                 // Field scala/runtime/ScalaRunTime$.MODULE$:Lscala/runtime/ScalaRunTime$;
         4: aload_1
         5: invokevirtual #35                 // Method scala/runtime/ScalaRunTime$.wrapIntArray:([I)Lscala/collection/immutable/ArraySeq;
         8: aload_2
         9: invokevirtual #38                 // Method withAnnotation:(Lscala/collection/immutable/Seq;Ljava/lang/String;)I
        12: ireturn
```

But this can't ever be called with varargs from Java, since while this is valid JVM bytecode, there is no equivalent Java code, where the varargs must always be last.

In fact, I'd argue that the annotation here can't be useful, and should be deprecated in scala

util-core/src/main/scala/com/twitter/util/StorageUnit.scala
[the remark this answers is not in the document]
yufangong: gotcha, I will only add the `.` then.

martijnhoekstra:
@yufangong Do you have an update how benchmarks are going? Still ongoing? Are there already any provisional "red" results I can start working on?

yufangong:
I am only running against 2.12, no obvious red flag so far
LocalBenchmark seems improved.
I will keep update results here:
https://gist.github.com/yufangong/61eeac9bc0765e913985e743c5300028

martijnhoekstra:
@yufangong I snuck sbt 1.3.0 RC3 in 
 in https://github.com/twitter/util/pull/248/commits/85fb0fdd64261e3dbb23eeb0d57b7797b03776c2  because RC2 has some annoying bugs.

I'm actually not too sure if sbt 1.3 should be part of this PR at all, or be factored into a different PR entirely (when 1.3.0 is fully released).

What's the status on your side?

yufangong:
@martijnhoekstra agreed that we should make it against a released version, I changed it to 1.2.8 and it seems working. Apologies that I was not actively working on this past week but going to get back on this branch next week.

martijnhoekstra:
@yufangong what's the current status on benchmarking and review? Is this good to go if the benches are good?

yufangong:
> @yufangong what's the current status on benchmarking and review? Is this good to go if the benches are good?

We have just upgraded Jackson version in source as part of the util cross-build effort. The most recent benchmark BrokerBenchmark seems a bit regression. I'll run again to see if thats valid.
I will move some questions in the review I cannot answer here. I will let you know if we need to take any actions, but so far it's going good.

util-app/src/main/scala/com/twitter/app/App.scala
```diff
@@ -313,7 +313,7 @@ trait App extends Closable with CloseAwaitably {
     deadline: Time
   ): Future[Unit] = {
     Future
-      .collectToTry(lastExits.asScala.toSeq.map(_.close(deadline)))
+      .collectToTry(lastExits.asScala.map(_.close(deadline)).toSeq)
```
yufangong: why does it matter where the toSeq is?

yufangong:
benchmarks are all clear, feel free to take a look at the result see if anything jumps out 
 https://gist.github.com/yufangong/61eeac9bc0765e913985e743c5300028

util-app/src/main/scala/com/twitter/app/App.scala
yufangong: ping

martijnhoekstra: I think it was part of a refactor that turned out wasn't needed, and this is a vestigial remainder of that refactor.

In other words, I don't think it matters, and pushed a commit to undo it

yufangong: thanks for clarifying.

yufangong:
@martijnhoekstra It is merged here https://github.com/twitter/util/commit/d5d20cc337710b07343eb09ecdab005968c97879! Yay! Thanks for your patience! Can you share us your strategy of this migration that we can pass to finagle/finatra/etc as well as the internal source repo? Are any automation tool or migration tool that is used and helpful? Thanks again!

martijnhoekstra:
This migration, I did with elbow grease: set the scala version, and fix all deprecation warnings and compiler errors.

For other repos, I'd make sure that you update all dependencies to a version that cross-builds, fix all deprecation warnings, then run the scalafix at https://github.com/scala/scala-collection-compat#collection213crosscompat, and take that as a starting point -- does the generated code make sense, or does it need an extra layer of shine. For example, in this repo, in many cases an `ArrayBuffer` was used to build a collection and return it as a (mutable) `Seq`. You can call `toSeq` at the end and call it good, but better is to use a `Builder` and return it's built results, which in many cases is `O(1)` rather than the `O(n)` that toSeq on the buffer would be. I don't think more automated fixes for that are feasible.

I intend to work my way towards Chill on 2.13 migration.

yufangong:
@martijnhoekstra Thanks so much! Closing this PR as resolved.

SethTisue:
ticket about publishing for 2.13.0: https://github.com/twitter/util/issues/253

PR #249, opened 2019-06-21
remove all RunWith

martijnhoekstra:
Problem

Superfluous calls to RunWith will require changes to the import for JUnitRunner.

Solution

Remove all RunWith and associated changes

Result

Code that won't need to be changed in the future to address deprecated alias for JUnitRunner


martijnhoekstra:
context at https://github.com/twitter/util/pull/248#discussion_r295534129

yufangong:
Thanks so much, this is merged at https://github.com/twitter/util/commit/ad9ccd08935e9516926c2ff86890c6783727b0e2.

Issue #245, opened 2019-04-10
Minor follow-ups from PR #242


kevinoliver:
Merging PR #242 fixed a broken window. Thanks! 

While merging this internally a few comments were brought up by folks. These were not blockers. Think of them as worth more discussion.

Roughly:
* In `get`, maybe call `Class.forName(className)` once and pass it in to `tryModuleField` and `tryMethod` to reduce overhead?
* In `getAll`, why do we parse `cls` from `info.className`, then do `get(info.className)` vs `get(cls)`? maybe worth reworking `get()` to take `Class`?
* In the test "GlobalFlag.getAll", could we change this to verify the existence of each of the global flags in the `MockClassLoader`?
* In the tests, can we add tests to verify that it doesn't load classes that aren't global flags? it is only checking the happy path today.

monkeyroar:
I'll look into it, can you assign this issue to me?

kevinoliver:
> I'll look into it, can you assign this issue to me?

awesome, thanks!

i must not know the github magic. you don't show up when i try to assign it to you. weird!

luciferous:
@monkeyroar all yours

Anurag870:
Created a PR for this.

yufangong:
This is fixed by https://github.com/twitter/util/pull/250, closing this. Thanks all.

PR #250, opened 2019-07-27
Fix for 245

Anurag870:
Making changes as per issue #245 which is follow up of #242

### DIFF
--- a/util-app/src/main/scala/com/twitter/app/GlobalFlag.scala
+++ b/util-app/src/main/scala/com/twitter/app/GlobalFlag.scala
@@ -120,10 +120,7 @@ object GlobalFlag {
     val className = if (!flagName.endsWith("$")) flagName + "$" else flagName
     try {
       get(Class.forName(className))
-    } catch {
-        case _: ClassNotFoundException | _: NoSuchFieldException | _: IllegalArgumentException =>
-          None
-      }
+    } catch { case _: ClassNotFoundException  => None }
   }
 
   private[app] def get(cls: Class[_]): Option[Flag[_]] = {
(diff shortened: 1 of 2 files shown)


Anurag870:
Signed CLA and linked email.

kevinoliver:
Thanks for the patch! It seems like this causes the util-app unit test to fail which you can see in the travis-ci link (https://api.travis-ci.org/v3/job/564437988/log.txt).

```
[info] - App: pass in bad args and expect usage *** FAILED ***
[info]   "Exception thrown in main on startup" did not contain "Error parsing flag "environment"" (AppTest.scala:74)
```

Please take a look, thanks.

Anurag870:
Thanks @kevinoliver, fixed it.

util-app/src/main/scala/com/twitter/app/GlobalFlag.scala
```diff
@@ -117,9 +117,18 @@ abstract class GlobalFlag[T] private[app] (
 object GlobalFlag {
 
   private[app] def get(flagName: String): Option[Flag[_]] = {
-    def tryMethod(className: String, methodName: String): Option[Flag[_]] =
+    val className = if (!flagName.endsWith("$")) flagName + "$" else flagName
+    try {
+      get(Class.forName(className))
+    } catch {
+        case _: ClassNotFoundException | _: NoSuchFieldException | _: IllegalArgumentException =>
```
yufangong: Do we need the NoSuchFieldException & IllegalArgumentException here?

util-app/src/main/scala/com/twitter/app/GlobalFlag.scala
```diff
@@ -130,9 +139,8 @@ object GlobalFlag {
           None
```
yufangong: Should we remove ClassNotFoundException here?

util-app/src/main/scala/com/twitter/app/GlobalFlag.scala
Anurag870: @yufangong Thanks, fixed it

yufangong:
@Anurag870 This is merged at https://github.com/twitter/util/commit/22b99087ade627f4e23404b12968e3609fe36f1f. Thanks for your contribution!

PR #251, opened 2019-07-29
App close race condition

gpoulin:
Problem

There's a race-condition in the close sequence of `com.twitter.app.App` when `close` doesn't meet the deadline. Normally, if the close sequence take longer than the graceful period, `closing` should be fulfilled by a `CloseException` due to [this](https://github.com/twitter/util/blob/util-19.7.0/util-app/src/main/scala/com/twitter/app/App.scala#L285) and [this](https://github.com/twitter/util/blob/util-19.7.0/util-app/src/main/scala/com/twitter/app/App.scala#L317) which is caught [here](https://github.com/twitter/util/blob/util-19.7.0/util-app/src/main/scala/com/twitter/app/App.scala#L71) and cause the JVM to be killed. However if this `Await.result` throw a `TimeoutException` before that, the app can be kept in a bad state.

Solution

Catch `TimeoutException` by `Await.result` and convert them into `CloseException`

Result

The App will always call `System.exit(1)` if `App#close` doesn't meet the deadline.

### DIFF
--- a/util-app/src/main/scala/com/twitter/app/App.scala
+++ b/util-app/src/main/scala/com/twitter/app/App.scala
@@ -97,8 +97,14 @@ trait App extends Closable with CloseAwaitably {
     // want to ensure "reason" is written before attempting to write any details
     System.err.flush()
     System.err.println(details)
-    Await.ready(close(), closeDeadline - Time.now)
-    kill()
+    try {
+      Await.ready(close(), closeDeadline - Time.now)
+    } catch {
+      case t: TimeoutException =>
+        System.err.println(s"Timeout while closing the App ${t.getMessage}")
+    } finally {
+      kill()
+    }
   }
 
   /**
@@ -377,8 +383,13 @@ trait App extends Closable with CloseAwaitably {
     close(defaultCloseGracePeriod)
 
     // The deadline to 'close' is advisory; we enforce it here.
-    if (!suppressGracefulShutdownErrors) Await.result(this, closeDeadline - Time.now)
-    else {
+    if (!suppressGracefulShutdownErrors) {
+      try {
+        Await.result(this, closeDeadline - Time.now)
+      } catch {
+        case t: TimeoutException => throw newCloseException(Seq(t))
+      }
+    } else {
       try { // even if we suppress shutdown errors, we give the resources time to close
         Await.ready(this, closeDeadline - Time.now)
       } catch {
--- a/util-app/src/main/scala/com/twitter/app/App.scala
+++ b/util-app/src/main/scala/com/twitter/app/App.scala
@@ -64,6 +64,10 @@ trait App extends Closable with CloseAwaitably {
    */
   protected def failfastOnFlagsNotParsed: Boolean = false
 
+
+  /** Forcibly exit the JVM without cleanup */
+  protected def kill(): Unit = System.exit(1)
+
   /** Exit on error with the given Throwable */
   protected def exitOnError(throwable: Throwable): Unit = {
     throwable.printStackTrace()
@@ -71,7 +75,7 @@ trait App extends Closable with CloseAwaitably {
       case _: CloseException =>
         // exception occurred while closing, do not attempt to close again
         System.err.println(throwable.getMessage)
-        System.exit(1)
+        kill()
       case _ =>
         exitOnError("Exception thrown in main on startup")
     }
@@ -94,7 +98,7 @@ trait App extends Closable with CloseAwaitably {
     System.err.flush()
     System.err.println(details)
     Await.ready(close(), closeDeadline - Time.now)
-    System.exit(1)
+    kill()
   }
 
   /**
--- a/util-app/src/test/scala/com/twitter/app/AppTest.scala
+++ b/util-app/src/test/scala/com/twitter/app/AppTest.scala
@@ -311,6 +311,21 @@ class AppTest extends FunSuite {
     assert(closed)
   }
 
+  test("App: exit even if the ShutdownTimer doesn't schedule timeout promptly") {
+    val app = new TestApp(() => ()) {
+     // We use a MockTimer to simulate jitter in the scheduling of timeout
+     override protected lazy val shutdownTimer = new MockTimer
+     def tick() = shutdownTimer.tick()
+    }
+
+    app.closeOnExit(Closable.make { _ => Future.never })
+
+    assert(!app.killed)
+    app.main(Array.empty)
+    app.tick()
+    assert(app.killed)
+  }
+
   test("App: exit functions properly capture non-fatal exceptions") {
     val app = new ErrorOnExitApp {
       def main(): Unit = {
--- a/util-app/src/test/scala/com/twitter/app/AppTest.scala
+++ b/util-app/src/test/scala/com/twitter/app/AppTest.scala
@@ -48,7 +48,7 @@ class AppTest extends FunSuite {
     test1.main(Array())
 
     assert(test1.reason.contains("Exception thrown in main on startup"))
-    assert(app.killed)
+    assert(test1.killed)
   }
 
   test("App: propagate underlying exception from fields in app") {
--- a/util-app/src/test/scala/com/twitter/app/AppTest.scala
+++ b/util-app/src/test/scala/com/twitter/app/AppTest.scala
@@ -311,18 +311,19 @@ class AppTest extends FunSuite {
     assert(closed)
   }
 
-  test("App: exit even if the ShutdownTimer doesn't schedule timeout promptly") {
+  test("App: exit even if `closing` is not fulfilled by `closeDeadline`") {
     val app = new TestApp(() => ()) {
-     // We use a MockTimer to simulate jitter in the scheduling of timeout
-     override protected lazy val shutdownTimer = new MockTimer
-     def tick() = shutdownTimer.tick()
+      // Timer that systematically miss its scheduling target by 500 milliseconds which could happen on a host under load
+      override lazy val shutdownTimer: Timer = new JavaTimer(isDaemon=true) {
+        override protected def scheduleOnce(when: Time)(f: => Unit): TimerTask =
+          super.scheduleOnce(when + 500.milliseconds)(f)
+      }
     }
 
     app.closeOnExit(Closable.make { _ => Future.never })
 
     assert(!app.killed)
     app.main(Array.empty)
-    app.tick()
     assert(app.killed)
   }
 
--- a/util-app/src/test/scala/com/twitter/app/AppTest.scala
+++ b/util-app/src/test/scala/com/twitter/app/AppTest.scala
@@ -9,8 +9,15 @@ import scala.language.reflectiveCalls
 
 class TestApp(f: () => Unit) extends App {
   var reason: Option[String] = None
+  var killed: Boolean = false
+
+  protected override def kill(): Unit = {
+    killed = true
+  }
+
   protected override def exitOnError(reason: String, details: => String): Unit = {
     this.reason = Some(reason)
+    super.exitOnError(reason, details)
   }
 
   def main(): Unit = f()
@@ -41,6 +48,7 @@ class AppTest extends FunSuite {
     test1.main(Array())
 
     assert(test1.reason.contains("Exception thrown in main on startup"))
+    assert(app.killed)
   }
 
   test("App: propagate underlying exception from fields in app") {
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -25,6 +25,12 @@ Runtime Behavior Changes
 
 * util: Upgrade to Jackson 2.9.9. ``PHAB_ID=D345969``
 
+Bug Fixes
+~~~~~~~~~
+
+* util-app: Fixed race-condition in `com.twitter.app.App` where the App could be kept running if
+  `App#close` is not fulfilled within the graceful period
+
 19.7.0
 ------
 


util-app/src/main/scala/com/twitter/app/App.scala
```diff
@@ -373,8 +383,13 @@ trait App extends Closable with CloseAwaitably {
...
-    else {
+    if (!suppressGracefulShutdownErrors) {
+      try {
+        Await.result(this, closeDeadline - Time.now)
+      } catch {
+        case t: TimeoutException => throw newCloseException(Seq(t))
+      }
+    } else {
       try { // even if we suppress shutdown errors, we give the resources time to close
         Await.ready(this, closeDeadline - Time.now)
```
gpoulin: I'm not sure what should be the behaviour here. Based on the comment on `suppressGracefulShutdownErrors` I would expect the App to still try to kill the JVM in case of error, simply with `System.exit(0)` instead of `System.exit(1)`.

util-app/src/test/scala/com/twitter/app/AppTest.scala
```diff
@@ -303,6 +311,21 @@ class AppTest extends FunSuite {
...
   }
 
+  test("App: exit even if the ShutdownTimer doesn't schedule timeout promptly") {
+    val app = new TestApp(() => ()) {
+     // We use a MockTimer to simulate jitter in the scheduling of timeout
+     override protected lazy val shutdownTimer = new MockTimer
+     def tick() = shutdownTimer.tick()
+    }
+
+    app.closeOnExit(Closable.make { _ => Future.never })
```
cacoco: Hey, thanks for the patch. I'm not 100% convinced by the premise here and and looking through, but this test is flawed. You have to wrap the `tick()` call in `Time.withCurrentTimeFrozen`, see the other tests here. We also test the close behavior in the presence of a `TimeoutException` with other tests here as well.

gpoulin: Thanks for looking into this. I think this simplified example might help to understand:

```scala
val closeDeadline: Time = ???
val closing = Future.never.by(shutdownTimer, closeDeadline, new CloseException)
Await.result(closing, closeDeadline - Time.now) 
```

This snippet could result in a `CloseException` if the timer fulfilled `closing` promptly or a `TimeoutException` otherwise. The `TimeoutException` case, while possible is unlikely since the timeout duration pass to will eventually be converted back to a deadline (by `java.util.concurrent.CountDownLatch`) but using `System.nanoTime` several operation later. However this case is still possible, we observed it in production and currently `App` doesn't handle it properly.

This unit-test doesn't really need to freeze the time, it just need to make sure that the timer doesn't schedule anything before the main function finish. I would argue that using `Time.withCurrentTimeFrozen` would be a little deceptive in that case, since we a looking at  a race-condition involving `Await.result` which rely on `java.util.concurrent.CountDownLatch` for is timeout so the time wouldn't be frozen for this timeout. However I agree that the current unit-test look shady and artificial, but I wasn't sure the best to trigger this race-condition reliably. I modified the unit-test to use a `Timer` that schedule everything 500ms late which is still subject to the same race-condition but way skewed in the other direction and it feel less artificial.

cacoco: Thanks for the information -- my main comment about using `Time.withCurrentTimeFrozen` is that you cannot properly mock time without using it.

We would still like to fully understand the issue and the impact of these changes. As you may be aware, `App` is the basis of all of our services and thus there is a very high bar for changes, especially when they deal with the lifecycle of the application, like close -- and definitely when the new logic is wired up to a `System.exit()` -- we would like to tread carefully, here. 

The initial reaction is a strong desire to not introduce a `kill` function in this manner and therefore I think we'd like to understand the issue and make sure we're actually addressing it and not a side-effect of something else that should or could be better handled elsewhere. 

To that end, your simple example has nothing to do with `App` and thus it is somewhat confusing as to why the solution would be in `App` (at least to me)? My point here is to make sure we are observing and testing the `App` API here if the solution is in `App`, since there is more that happens in `close` than the provided snippet.

To make sure I understand the gist of the problem correctly, you've observed that a `TimeoutException` can escape when closing the application instead of the expected `CloseException` and since the `TimeoutException` is not handled, bad things can happen -- is that accurate? I want to be clear that we are seeing this behavior through the normal flow of the `App#close` method. Thanks.

gpoulin: The `App` currently `Await` on itself at the end of the `main` [ref](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L376). If one of the `Closable` is never fulfilled, this `Await` can result in a `CloseException` or a `TimeoutException` due to a race-condition within the `App`. Currently only the `CloseException` is handled [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L71) and ultimately result in a `System.exit(1)`. The `TimeoutException` would result in calling `exitOnError(reason: String, details: => String)` which will itself throw a `TimeException` [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L96). There's no way to handle that downstream apart from overriding `exitOnError` which doesn't seem like a good idea.

The modified unit-test show the issue and simply call `App#main`. The only "strange" part is the `Timer` that schedule with a delay, but that can happen on the JVM since according to the [doc](https://docs.oracle.com/javase/8/docs/api/java/util/Timer.html) `This class does not offer real-time guarantees: it schedules tasks using the Object.wait(long) method.` If you are still not convinced by this unit-test, [this](https://github.com/twitter/util/compare/develop...gpoulin:simple_timeout?expand=1#diff-9f877f117b521ff3cfd79847c7fcf534R314) simple one fail about half the time when I try it locally, which also mean it succeed about half the time which is not great for a unit-test.

The addition of the `kill` method is only needed for the unit-test, as currently this race condition will cause `exitOnError(reason: String)` to be called (which is what other unit-test check), but this method will throw a `TimeoutException` and never reach `System.exit(1)`. I would be happy to do away with this `kill` method if you know of a way that I can unit-test `exitOnError(reason: String)`.

gpoulin: To clarify, by "simplified example", I meant a simplify version of the race-condition that exist in `App`. This example, is basically what `App` does if you remove the logic that doesn't concern this race-condition.

`val closeDeadline: Time = ???` -> [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L208)
`val closing = Future.never.by(shutdownTimer, closeDeadline, new CloseException)` -> [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L280)
`Await.result(closing, closeDeadline - Time.now)` -> [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L376)

util-app/src/main/scala/com/twitter/app/App.scala
```diff
@@ -373,8 +383,13 @@ trait App extends Closable with CloseAwaitably {
     close(defaultCloseGracePeriod)
 
     // The deadline to 'close' is advisory; we enforce it here.
-    if (!suppressGracefulShutdownErrors) Await.result(this, closeDeadline - Time.now)
-    else {
+    if (!suppressGracefulShutdownErrors) {
```
cacoco: If seems we could replace all of this with:

```
    // The deadline to 'close' is advisory; we enforce it here.
    try {
      Await.result(this, closeDeadline - Time.now)
    } catch {
      case NonFatal(exc) =>
        if (!suppressGracefulShutdownErrors) {
          exitOnError(newCloseException(Seq(exc)))
        } else Unit
    }
```
I think the issue continues to be a way to test this.

cacoco: The point of the `suppressGracefulShutdownErrors` was to make sure the process exited with a successful exit code (0) and that any error here did not bubble up.

gpoulin: I understand the idea of `suppressGracefulShutdownErrors`, however it won't call `System.exit(0)` and only rely on the `main` finishing. So if there is a non-daemon thread still running because of a closing exception, the app will never be closed. That's what I meant [here](https://github.com/twitter/util/pull/251#discussion_r308465982).

You say that we need a way to test this. Is there anything wrong with the jittered timer approach? If so what are you suggesting?

cacoco: The current machinery is meant to throw an exception due to the result of `Await.result` or to try to let any draining happen with the `Await.ready` (which is just checking if the `this` Future is eventually satisfied). The semantics of the two are subtly different.

In the first case, it is expected that any exception bubbles up the stack and is caught and handled by anything that cares to perform a clean shutdown -- otherwise it uncaught and the JVM exits. Whereas the latter case is purposely intended to not bubble up any errors on closing in this manner as it is expected that the orchestration layer is tearing things down anyway and, yes, relies on the main simply exiting.

There's a lot of machinery built on this and how this works, so changing things here is generally not simple. Even my example above fails for lots of reasons under testing.

It sounds like you're in a situation where you have started daemon threads and an exception happens when trying to close them which prevents the parent process from shutting down. I don't think there is much we can do here in the App code to help in this situation. Trying to signal that the app should terminate the JVM is unfortunately not possible in a non API breaking way currently.

My suggestion would be to create a wrapper process which calls to `App.main` in a try/catch which then can call System.exit for you.

cacoco: Another suggestion is to introduce another API to signal that you explicitly want to "exit" on error. We use `App` primarily through `TwitterServer` internally and in the semantics of that usage are to call `close()` manually on the server as it is blocking in the call to main in `App` [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L363). We expect the main to then exit accordingly without terminating the JVM with a `System.exit` call.

We could introduce a boolean, say, `def exitOnError: Boolean = false` which could be overridden to `true` in your application. It would trigger the `exitOnError(throwable)` when an exception is encountered and application has signaled this should exit the JVM.

Note: I'm sure we can re-use `exitOnError` like this but it's as good a name for the variable as  any.

cacoco: I have to admit, I am also still confused about the use case you are trying to solve. The call to `nonExitingMain` is done by `main` and is wrapped in a try/catch [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L331-L340). In the patched code, you are handling a `TimeoutException` which should be emitted from the `nonExitingMain` and intercepted by the case in the try/catch in main. Which would then terminate through one of two code paths: either directly [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L71-L74) due to getting a `CloseException` or by trying to wait a little longer for the future to complete and then bailing [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L91-L97).

Based on your patch, I'm not sure why you'd need to touch the code in the `nonExitingMain` which is maybe why I'm confused about the situation you're trying to handle. I could see trying to patch the `Await.ready` [here](https://github.com/twitter/util/blob/develop/util-app/src/main/scala/com/twitter/app/App.scala#L96) if things are hung to make sure it continues to the `System.exit`. But I would think that if you're trying to get the process to exit due to a `TimeoutException` on either the `Await.result` or `Await.ready` calls in the `nonExitingMain`, it does?

Also, I meant to answer you about the "jittered timer approach" -- yes, my concerns would be that test would be flakey in the sense it would not be deterministic.

cacoco:
@gpoulin can you take a look at the changes in b17297c731d7f693f0284a2d0d11e23635eff7e2 and see if they address your concerns? Thanks!

gpoulin:
@cacoco I think it address the main concern of this PR, however it seems strange/fragile that instead on consistently throwing the same exception in case of timeout we catch the two exception in two different places. It introduce slight behaviour difference base on which timeout win, but it likely doesn't really matter much since it's on `print` statements.

However, I'm not too sure about this change https://github.com/twitter/util/commit/b17297c731d7f693f0284a2d0d11e23635eff7e2#diff-f53f4d0030117b43334883c0dedd9d24R387 . It introduce a non-deterministic interface based on which timeout win. If the timer win, the exception would be catched and ignored but if `Await.ready` win, the a `TimeoutException` will be catched in main and `exitOnError` would be called eventually calling `System.exit(1)`.

cacoco:
@gpoulin there are definitely two different cases being handled here and it is intentional and I want to make sure it's clear that there is a difference between `Await.ready` and `Await.result`. I think I disagree with the idea that this is non-deterministic since as I said we have two different modes of operation we're trying to handle and I believe each is handled properly, but I may be wrong.

As I mentioned, `Await.result` and `Await.ready` are different. `Await.result` returns the result of the awaited Future. In this case, it is `this` or the App closing. This is such that we can capture the results of trying to close all of the registered exits. This will throw a `TimeoutException`. This is caught by the `try/catch` executing the `nonExitingMain` function in `main`. And `exitOnError` is called which will shut down the JVM properly.

In the other case, we do not care about the result of the awaited Future, merely we want to know that is was eventually satisfied. This is what `Await.ready` does, it just blocks until the awaited Future is satisfied. This code incorrectly didn't care about a possibly thrown `TimeoutException` as it swallowed all exceptions. Thus the changes make sure we propagate a `TimeoutException` like in the first case which is then handled by the `try/catch` executing the `nonExitingMain` function in `main`. 

The last change is to make sure we address the same problem in `exitOnError` itself. Hopefully, that helps explain.

Thanks!

gpoulin:
To be clear part that is non-deterministic is that `Await.ready` might not throw a `TimeoutException` if there's a timeout since the future might be fulfilled by the shutdown timer before. Before the change it was deterministic because the all error was ignore, but now the same issue can result in a `TimeoutException` throw by `Await.ready` or a `CloseException` which is ignored by `Await.ready`.

cacoco:
Closing as answered.

PR #252, opened 2019-08-06
Make flatMap on AsyncStream stack-safe

travisbrown:
## Problem

The recent `com.twitter.util.BypassScheduler` optimization breaks stack-safety for monadic recursion for `AsyncStream`. For example, on 19.7.0 or 19.8.0 (or 19.6.0 with `-Dcom.twitter.util.BypassScheduler=true`) the following will overflow, while it used to work just fine:

```scala
import com.twitter.concurrent.AsyncStream

def deep(i: Int): AsyncStream[Int] =
  if (i == 0) AsyncStream.of(i) else AsyncStream.of(i).flatMap(i => deep(i - 1))

com.twitter.util.Await.result(deep(10000).toSeq)
```

## Solution

After some discussion we decided the best solution would be to replace `Future#map` with `Future#flatMap` in the definition of `AsyncStream`'s `flatMap` in order to make it stack-safe again.

~~I've added a new `tailRecM` method to the `AsyncStream` companion object that provides stack-safe monadic recursion via a (private) `stackSafeFlatMap` method that uses `flatMap`s on `Future` instead of `map`. I've also added a test and verified that it fails with the naive `flatMap`-based `tailRecM` implementation.~~

~~The name `tailRecM` is pretty standard: it's used in Cats and in other libraries and languages.~~

## Result

`flatMap` can be used recursively again without worrying about stack overflows.

### DIFF
--- a/util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
+++ b/util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
@@ -217,13 +217,18 @@ sealed abstract class AsyncStream[+A] {
   /**
    * Map a function `f` over the elements in this stream and concatenate the
    * results.
+   *
+   * @note We use `flatMap` on `Future` instead of `map` to maintain
+   * stack-safety for monadic recursion.
    */
   def flatMap[B](f: A => AsyncStream[B]): AsyncStream[B] =
     this match {
       case Empty => empty
-      case FromFuture(fa) => Embed(fa.map(f))
-      case Cons(fa, more) => Embed(fa.map(f)) ++ more().flatMap(f)
-      case Embed(fas) => Embed(fas.map(_.flatMap(f)))
+      case FromFuture(fa) => Embed(fa.flatMap(a => Future.value(f(a))))
+      case Cons(fa, more) =>
+        Embed(fa.flatMap(a => Future.value(f(a)))) ++ more().flatMap(f)
+      case Embed(fas) =>
+        Embed(fas.flatMap(as => Future.value(as.flatMap(f))))
     }
 
   /**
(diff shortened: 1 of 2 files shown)


util-core/src/main/scala/com/twitter/concurrent/AsyncStream.scala
```diff
@@ -672,6 +672,34 @@ object AsyncStream {
...
+   */
+  private def stackSafeFlatMap[A, B](as: AsyncStream[A])(f: A => AsyncStream[B]): AsyncStream[B] =
+    as match {
+      case Empty => empty
+      case FromFuture(fa) => Embed(fa.flatMap(a => Future.value(f(a))))
+      case Cons(fa, more) =>
+        Embed(fa.flatMap(a => Future.value(f(a)))) ++ stackSafeFlatMap(more())(f)
+      case Embed(fas) =>
+        Embed(fas.flatMap(as => Future.value(stackSafeFlatMap(as)(f))))
+    }
```
luciferous: Was chatting with @travisbrown on Gitter and I think it's worth considering fixing flatMap (i.e. just replace flatMap with the implementation of stackSafeFlatMap above).

The performance gains of bypass scheduler apply to Future, but they weren't intended to break stack-safety of AsyncStream.flatMap. `flatMap` is commonly used for recursive constructs, we should make it safe to use that way.

travisbrown: @luciferous Updated the PR, thanks!

luciferous:
@travisbrown can update commit the PR description?

luciferous:
Anyone know why the CI isn't passing?

yufangong:
> Anyone know why the CI isn't passing?
It is the CumulativeGaugeTest being flaky, not related to the change here.

luciferous:
Closed by 116988077ed1452c749286087195ce984262083a.

PR #257, opened 2019-10-02
Unable to stop CollectClosables thread. Resolve #758

DanielCharczynski:
module-name: util-core

Problem

No possibility to stop _CollectClosables_ thread.
Problem described here https://github.com/twitter/finagle/issues/758

Solution

Created public method `Closable.stopCollectClosablesThread` that stops _CollectClosables_ thread

Result

New API that makes possible to stop _CollectClosables_ thread.

### DIFF
--- a/util-core/src/main/scala/com/twitter/util/Closable.scala
+++ b/util-core/src/main/scala/com/twitter/util/Closable.scala
@@ -45,6 +45,14 @@ abstract class AbstractClosable extends Closable
  */
 object Closable {
   private[this] val logger = Logger.getLogger("")
+  private[this] var collectorThreadEnabled = true
+  
+  
+  /** Stop CollectClosables thread. */
+  def stopCollectClosablesThread() {
+	collectorThreadEnabled = false
+	collectorThread.interrupt()
+  }
 
   /** Provide Java access to the [[com.twitter.util.Closable]] mixin. */
   def close(o: AnyRef): Future[Unit] = o match {
@@ -161,7 +169,7 @@ object Closable {
 
   private val collectorThread = new Thread("CollectClosables") {
     override def run(): Unit = {
-      while (true) {
+      while (collectorThreadEnabled) {
         try {
           val ref = refq.remove()
           val closable = refs.synchronized(refs.remove(ref))
--- a/util-core/src/main/scala/com/twitter/util/Closable.scala
+++ b/util-core/src/main/scala/com/twitter/util/Closable.scala
@@ -6,6 +6,7 @@ import java.util.concurrent.atomic.AtomicReference
 import java.util.logging.{Level, Logger}
 import scala.annotation.{tailrec, varargs}
 import scala.util.control.{NonFatal => NF}
+import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * Closable is a mixin trait to describe a closable ``resource``.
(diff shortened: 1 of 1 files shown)


DanielCharczynski:
I have signed Contributor License Agreement five times but it looks like it is no visible here....

mosesn:
This patch seems reasonable to me.  Can you follow the guidelines in CONTRIBUTING.md about your commit message / github PR description? LGTM otherwise.

DanielCharczynski:
Done

mosesn:
Great, thanks!  We'll make sure it passes our internal tests too and then merge it in.

util-core/src/main/scala/com/twitter/util/Closable.scala
```diff
@@ -45,6 +45,14 @@ abstract class AbstractClosable extends Closable
  */
 object Closable {
   private[this] val logger = Logger.getLogger("")
+  private[this] var collectorThreadEnabled = true
```
ryanoneill: What do you think about making this volatile? I would expect the callers of `stopCollectClosablesThread` to not be the "CollectClosables" thread itself, and so there may be memory visibility issues here.

util-core/src/main/scala/com/twitter/util/Closable.scala
```diff
@@ -45,6 +45,14 @@ abstract class AbstractClosable extends Closable
...
+  
+  
+  /** Stop CollectClosables thread. */
+  def stopCollectClosablesThread() {
+	collectorThreadEnabled = false
+	collectorThread.interrupt()
```
ryanoneill: Calling `interrupt` can throw a `SecurityException`. It _shouldn't_ in our case, but perhaps we should catch the exception here, log and just eat it.

util-core/src/main/scala/com/twitter/util/Closable.scala
```diff
@@ -45,6 +45,14 @@ abstract class AbstractClosable extends Closable
  */
 object Closable {
   private[this] val logger = Logger.getLogger("")
+  private[this] var collectorThreadEnabled = true
+  
+  
+  /** Stop CollectClosables thread. */
+  def stopCollectClosablesThread() {
+	collectorThreadEnabled = false
```
ryanoneill: We should perhaps protect against calling this multiple times.

util-core/src/main/scala/com/twitter/util/Closable.scala
ryanoneill: Could also make it an AtomicBoolean.

DanielCharczynski: AtomicBoolean already uses 'volatile' variable

util-core/src/main/scala/com/twitter/util/Closable.scala
DanielCharczynski: OK

util-core/src/main/scala/com/twitter/util/Closable.scala
DanielCharczynski: Done

jyanJing:
Thank you so much for working on it @DanielCharczynski , the change is merged in [a826099](https://github.com/twitter/util/commit/a8260998518a9efc32ece0c5114694674c63072c)

PR #260, opened 2019-12-15
util-stats: Limit the max stats in InMemoryStatsReceiver

takezoe:
## Problem

`InMemoryStatsReceiver` keeps all added stats in memory so it consumes the heap infinitely.

## Solution

This pull request enables to limit the maximum number of stats stored in memory. The default is no limitation as before.

### DIFF
--- a/util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
+++ b/util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
@@ -118,7 +118,11 @@ class InMemoryStatsReceiver(maxStats: Int) extends StatsReceiver with WithHistog
 
       def add(value: Float): Unit = stats.synchronized {
         val oldValue = apply()
-        stats(schema.metricBuilder.name) = oldValue.takeRight(maxStats) :+ value
+        if(maxStats > 0) {
+          stats(schema.metricBuilder.name) = oldValue.takeRight(maxStats) :+ value
+        } else {
+          stats(schema.metricBuilder.name) = oldValue :+ value
+        }
       }
       def apply(): Seq[Float] = stats.getOrElse(schema.metricBuilder.name, Seq.empty)
 
--- a/util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
+++ b/util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
@@ -50,8 +50,10 @@ object InMemoryStatsReceiver {
  *
  * @param maxStats the maximum number of stats, zero or negative value means unlimited (default is unlimited)
  **/
-class InMemoryStatsReceiver(maxStats: Int = 0) extends StatsReceiver with WithHistogramDetails {
+class InMemoryStatsReceiver(maxStats: Int) extends StatsReceiver with WithHistogramDetails {
   import InMemoryStatsReceiver._
+
+  def this() = this(0)
 
   def repr: InMemoryStatsReceiver = this
 
@@ -116,16 +118,7 @@ class InMemoryStatsReceiver(maxStats: Int = 0) extends StatsReceiver with WithHi
 
       def add(value: Float): Unit = stats.synchronized {
         val oldValue = apply()
-
-        // Limit the maximum number of stats as 10000000
-        val length = oldValue.length
-        val tailValue = if (maxStats <= 0 || length < maxStats) {
-          oldValue
-        } else {
-          oldValue.drop(length - maxStats + 1)
-        }
-
-        stats(schema.metricBuilder.name) = tailValue :+ value
+        stats(schema.metricBuilder.name) = oldValue.takeRight(maxStats) :+ value
       }
       def apply(): Seq[Float] = stats.getOrElse(schema.metricBuilder.name, Seq.empty)
 
--- a/util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
+++ b/util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
@@ -47,8 +47,10 @@ object InMemoryStatsReceiver {
  * assert(isr.counters(Seq("a", "b", "foo") == 0)) // ok
  * assert(isr.counters(Seq("a", "b", "bar") == 0)) // fail
  * }}}
+ *
+ * @param maxStats the maximum number of stats, zero or negative value means unlimited (default is unlimited)
  **/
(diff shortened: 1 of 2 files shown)


mgodave:
@takezoe thanks for the contribution! Would it be possible to add some tests around the new functionality.

util-stats/src/test/java/com/twitter/finagle/stats/StatsReceiverCompilationTest.java
```diff
@@ -23,7 +23,7 @@ public void testNullStatsReceiver() {
 
   @Test
   public void testStatMethods() {
-    InMemoryStatsReceiver sr = new InMemoryStatsReceiver();
+    InMemoryStatsReceiver sr = new InMemoryStatsReceiver(0);
```
mgodave: are these changes necessary since the default argument is zero?

util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
```diff
@@ -47,8 +47,10 @@ object InMemoryStatsReceiver {
  * assert(isr.counters(Seq("a", "b", "foo") == 0)) // ok
  * assert(isr.counters(Seq("a", "b", "bar") == 0)) // fail
  * }}}
+ *
+ * @param maxStats the maximum number of stats, zero or negative value means unlimited (default is unlimited)
  **/
-class InMemoryStatsReceiver extends StatsReceiver with WithHistogramDetails {
+class InMemoryStatsReceiver(maxStats: Int = 0) extends StatsReceiver with WithHistogramDetails {
```
mosesn: instead of using a default arg, please overload the constructor so it's java friendly (java doesn't know how to deal with default args).

also, would it more clearly communicate our intent to pass this as an Option, and require that it's > 0 if specified?

util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
```diff
@@ -114,7 +116,16 @@ class InMemoryStatsReceiver extends StatsReceiver with WithHistogramDetails {
 
       def add(value: Float): Unit = stats.synchronized {
         val oldValue = apply()
-        stats(schema.metricBuilder.name) = oldValue :+ value
+
+        // Limit the maximum number of stats as 10000000
```
mosesn: where does `10000000` come from?

util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
```diff
@@ -114,7 +116,16 @@ class InMemoryStatsReceiver extends StatsReceiver with WithHistogramDetails {
...
+        val length = oldValue.length
+        val tailValue = if (maxStats <= 0 || length < maxStats) {
```
mosesn: ffti: `stats(schema.metricBuilder.name) = oldValue.takeRight(maxStats) :+ value`

util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
takezoe: Oh, it came from my experiment code. Just removed thet comment.

util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
takezoe: Thanks! Updated.

util-stats/src/test/java/com/twitter/finagle/stats/StatsReceiverCompilationTest.java
takezoe: They are necessary in Java code.

takezoe:
Thanks for your review!

We are using our own in-memory stats receiver implementation on Finagle in production and we just thought we can use twitter-util's `InMemoryStatsReceiver` instead of ours if it supports expiration of recorded stats.

If this contradicts the design intention of `InMemoryStatsReceiver`, feel free to reject this pull request. We can understand your decision and we can continue to use our own in-memory solution.

util-stats/src/main/scala/com/twitter/finagle/stats/InMemoryStatsReceiver.scala
takezoe: Yeah, so that we don't need to give a parameter in Java code.

mosesn:
@takezoe how about using `finagle-stats` instead?  Do you need the precision that InMemorySR gives you?  Can you tell me more about what you're using it for?  We have a pretty decent production-friendly metrics implementation in `finagle-stats-core` that can extended for many other uses, or if you're just polling the metrics you can integrate with `TwitterServer` to export them from `finagle-stats` pretty automatically.

mosesn:
Also, just fair warning–we might be a bit less responsive for the next couple of weeks, much of the team will be out for the holidays, but please keep commenting on the ticket and we'll try to get back to you when we can!

takezoe:
Our use case is that exposing Finagle metrics via JMX. We have a common framework for exposing JMX metrics and currently, out JMX frameworks retrieves Finagle metrics through a modified version of `InMemoryStatsReceiver`.

I looked through finagle-stats-core roughly. Maybe we can use it instead.

takezoe:
One thing I want to know about `MetricsStatsReceiver` which is provided by finagle-stats-core is how can we read metrics from `Metrics` although it's a package private.
https://github.com/twitter/finagle/blob/d42c87a97f25573759c1a6f605d0d2265f76619c/finagle-stats-core/src/main/scala/com/twitter/finagle/stats/Metrics.scala#L109

It might be a solution for JMX MBean to retrieves metrics from `/admin/metrics.json` which is served by `MetricsExporer`, but it's obviously not the best way. Rather, it might be better to implement our own receiver?

mosesn:
@takezoe yeah, I think it might be good to change `MetricsStore` and `MetricsView` to be public, change Metrics.createDetached to return a `MetricsStore with MetricsView` and then you can use Metrics.createDetached to make a nice backend for exporting to JMX.  Alternatively, if you're using codahale's metrics, someone built a connector called [finagle-metrics](https://github.com/rlazoti/finagle-metrics) a few years ago.  It's not up to date, but you could probably fork it and dust it off if that happens to be your exact use case (I know they can export to JMX, so that's just a shot in the dark).

I think that will be substantially higher performance than InMemoryStatsReceiver, which is intentionally naive.

takezoe:
Thanks! Since there are some better ways, I close this pull request and will try them.

PR #261, opened 2020-01-23
util-core: remove early initializer for TimeLike

martijnhoekstra:
Problem

Early initializers are deprecated and will be removed in a future version of scala.

Solution

Replace `TimeLike` trait with abstract class, and put `ops` in a constructor parameter instead of it being a value that needs to be early-initialized

Result

No more early initializers for future compatibility.

### DIFF
--- a/util-core/src/main/scala/com/twitter/util/Duration.scala
+++ b/util-core/src/main/scala/com/twitter/util/Duration.scala
@@ -304,7 +304,7 @@ private[util] object DurationBox {
  * that are truly infinite; for example the absence of a timeout.
  */
 sealed class Duration private[util] (protected val nanos: Long) extends TimeLike[Duration] with Serializable {
-  override def ops: TimeLikeOps[Duration] = Duration
+  override val ops: TimeLikeOps[Duration] = Duration
   import Duration._
 
   def inNanoseconds: Long = nanos
--- a/util-core/src/main/scala/com/twitter/util/Duration.scala
+++ b/util-core/src/main/scala/com/twitter/util/Duration.scala
@@ -304,10 +304,8 @@ private[util] object DurationBox {
  * their arithmetic follows. This is useful for representing durations
  * that are truly infinite; for example the absence of a timeout.
  */
-sealed class Duration private[util] (protected val nanos: Long) extends {
-  protected val ops: Duration.type = Duration
-} with TimeLike[Duration] with Serializable {
-  import ops._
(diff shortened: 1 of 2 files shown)


martijnhoekstra:
These test failures look spurious to me.

bryce-anderson:
@martijnhoekstra, I have bad news: this fails internally. We're using json4s and some code in there is reflectively searching for constructor parameter types, and ends up in a recursive loop and stack overflow. I suspect it's a flaw in json4s, but regardless it's a blocker for this patch. There is probably a pretty simple way forward though: what if instead of making `TimeLike` an abstract class with a constructor arg you left it a trait and made `ops` a `def`? You can't use the `import ops._` but it should be easy enough to just add the `ops.Whatever` to all the use cases. That will all but certainly avoid the SO and has the benefit of preserving the API since we keep `TimeLike` a trait instead of an abstract class.

martijnhoekstra:
I think I tried that first, but `TimeLike` itself uses some of the imported members. I'll see what was going on with that (and also see if I can find out what's going on with the stackoverflow in json4s, that probably deserves an issue on that in its own right)

bryce-anderson:
@martijnhoekstra, it looks like all the uses of `ops._` in `TimeLike` are in methods so there shouldn't be any issue with constructor ordering if `ops` becomes a `def`.

martijnhoekstra:
That seems to work. Could you check internally @bryce-anderson ?

martijnhoekstra:
If the imported parts are only used in defs, can there be an initialization order problem at all? I'm only seeing the usual windows-related errors when I run the testsuite.

bryce-anderson:
I'll run the full suite internally and let you know but I spot checked one of the tests that was failing and it works on my local machine. 😄

martijnhoekstra:
I meant if it's only used in defs, then we can make it a val, and eliminate the early initializers with just not doing that. I pushed a commit that does that. If that fails, the def approach should work.

bryce-anderson:
Tbh, I kind of liked it as a `def`: there isn't any advantage to the `val` as far as I can tell and it adds another pointer field to the `Time` and `Duration` types so making it a def potentially results in reduced allocations.

martijnhoekstra:
Would you like me to revert to that version? I have no strong feelings on the issue, so you could also just merge whichever version makes the most sense to you.

bryce-anderson:
Closed via ccfb944.

PR #262, opened 2020-01-23
util: replace deprecated JavaConverters

martijnhoekstra:
. Use scala.jdk.CollectionConverters instead

on 2.11.x and 2.12.x this is provided by the compat library

Problem

scala.collection.JavaConverters is deprecated

Solution

While it's unclear it's possible deprecated APIs will be removed before scala 4, replace them with scala.jdk.CollectionCompat anyway, even if only to get fewer warnings in the build.

Result

Fewer warnings.


bryce-anderson:
Merged as d8cc723962da36f6fe844f0d4a6d8dd42e9efbac. Thanks again!

PR #263, opened 2020-01-29
add missing type ascription to Logger's apply method

mtkp:
Problem

when logging strings that sometimes contain "%" (e.g. URLs with escaped characters), if i also log an exception, i might get a `java.util.UnknownFormatConversionException`. this only happens with the `Logger`'s `apply` method.

Solution

added a missing type ascription to the `apply` method, so it sends the correct items sequence to the `log` method.

Result

fixes added test case (using `apply` with strings that have %)


bryce-anderson:
Merged as c875d359d6505a6979258f9fd596fa05ebd3c792. Thanks!

mtkp:
thanks @bryce-anderson !

PR #279, opened 2020-06-26
Add integration with new async/await support in 2.12.12+ and 2.13.3+

olafurpg:
## Problem

Currently, the primary syntax to compose Twitter futures is using `map/flatMap` syntax. This syntax may feel verbose and unfamiliar to developers who have worked with other languages such as JavaScript or C# that support async/await.

## Solution

This PR adds async/await support for `com.twitter.util.Future`. This functionality is
only available in the newly released 2.13.3 (notes https://github.com/scala/scala/releases/tag/v2.13.3). It will also become available in the upcoming 2.12.12 release.

## Result

Users can write code like this now
```scala
val c = async {
  val a = Future.value(10)
  val b = Future.value(5)
  await(a) + await(b)
}
assert(c.poll == Some(Return(15)))
```

The equivalent way to to write the same code above with the current syntax is 
```scala
  val c = for {
    a <- Future.value(10)
    b <- Future.value(5)
  } yield a + b
```

## Notes

The code in this PR is adapted from the code in https://github.com/scala/scala-async/pull/237

To try out this PR locally run the following commands
```
$ sbt
> ++2.13.3
> util-core/testOnly com.twitter.util.AsyncTest
```


olafurpg:
For the record, I'm not very familiar with Twitter futures or the new -Xasync flag. This PR is just an experiment so I'm converting it into a draft.

Issue #266, opened 2020-02-21
Please reconsider #235


sjoerd-vogel:
Please reconsider #235 as it is required for jdk11 support.

### Expected behavior

https://twitter.github.io/finagle/guide/changelog.html mentions initial jdk11 support for 19.11.0 onwards. Hence I expect jdk11 support.

### Actual behavior

As far as I can tell issue #235 has not yet been fixed. It results in warnings in jdk11.

### Steps to reproduce the behavior

See issue #235 for this.

cacoco:
@sjoerd-vogel Hi there, thanks for the issue! Reading through the previous ticket, I don't think the intention was that we are done here -- it's just that we don't yet officially support anything other than JDK8 and when we move to full support of JDK11 we will tackle this (it should be within the next few quarters). Since it only results in warnings and not an error we're waiting until the move to address. Hop that helps. Thanks!

ComFreek:
Are there any news on this issue? JDK 8 has been long superseeded by newer versions.

ComFreek:
@cacoco Is there any update on this? :smile:

cacoco:
@ComFreek hey there! We run a very large codebase and are still in the process of certifying the move to JDK11. I don't have a current ETA here but the work is on-going. Hope that helps. Thanks!

sjoerd-vogel:
I gave this a shot. 
https://github.com/twitter/util/pull/284

felixbr:
So, [JDK8 is EoL end of March 2022](https://endoflife.date/java). Also there are several known bugs in JDK8 which would be fixed in JDK11+.

I suspect Twitter still doesn't support anything beyond JDK8 officially for their codebase but maybe we can at least try and support JDK11 for twitter-util, twitter-server, and finagle (= open-source libraries)?

When looking at the CI-config [here](https://github.com/twitter/util/blob/develop/.github/workflows/continuous-integration.yml#L25) it looks like the build is already tested for JDK11. In my mind this should be enough to accept PRs which try to improve JDK11 support as long as they don't break JDK8 support.

Am I missing something?

PR #284, opened 2020-11-14
java 11 reflect warning fixes

sjoerd-vogel:
Problem

See https://github.com/twitter/util/issues/266, the current code throws warnings under jdk11

Solution

In jdk11 it is possible to avoid the illegal reflection by simply calling getters that now exist. I tried to stay java 8 backwards compatible by catching a possible NoSuchMethodError. The new code _does_ need to be compiled using jdk11 however. Also its kinda hard to properly test this against jdk8.

Result

The warnings disappear in jdk11. Some additional (manual?) tests are required to check the jdk8 backwards compatibility.

### DIFF
--- a/util-jvm/src/main/scala/com/twitter/jvm/Hotspot.scala
+++ b/util-jvm/src/main/scala/com/twitter/jvm/Hotspot.scala
@@ -35,11 +35,18 @@ class Hotspot extends Jvm {
     ObjectName.getInstance("com.sun.management:type=HotSpotDiagnostic")
 
   private[this] val jvm: VMManagement = {
-    val fld = Class
-      .forName("sun.management.ManagementFactoryHelper")
-      .getDeclaredField("jvm")
-    fld.setAccessible(true)
-    fld.get(null).asInstanceOf[VMManagement]
+    try {
+      sun.management.ManagementFactoryHelper
+        .getVMManagement
+        .asInstanceOf[VMManagement]
+    } catch {
+      case _: NoSuchMethodError => {
+        val fld = Class.forName("sun.management.ManagementFactoryHelper")
+          .getDeclaredField("jvm")
+        fld.setAccessible(true)
+        fld.get(null).asInstanceOf[VMManagement]
+      }
+    }
   }
 
   private[this] def opt(name: String) =
@@ -150,7 +157,7 @@ class Hotspot extends Jvm {
     )
   }
 
-  private[this] object NilSafepointBean {
+  private[this] object NilSafepointBean extends SafepointBean {
     def getSafepointSyncTime = 0L
     def getTotalSafepointTime = 0L
     def getSafepointCount = 0L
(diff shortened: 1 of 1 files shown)


util-jvm/src/main/scala/com/twitter/jvm/Hotspot.scala
```diff
@@ -159,17 +166,10 @@ class Hotspot extends Jvm {
...
-      .invoke(null)
+    val runtimeBean = sun.management.ManagementFactoryHelper.getHotspotRuntimeMBean
 
     def asSafepointBean(x: AnyRef) = {
-      x.asInstanceOf[{
-        def getSafepointSyncTime: Long
-        def getTotalSafepointTime: Long
-        def getSafepointCount: Long
-      }]
+      x.asInstanceOf[sun.management.HotspotRuntimeMBean]
```
sjoerd-vogel, line 172: NB: I changed this because the original cast to an anonymous trait depended on reflection and caused a similar warning. This interface has the same methods but also contains an additional method `List<Counter> getInternalRuntimeCOunters()` If we want to hide this we might need a wrapper class.

util-jvm/src/main/scala/com/twitter/jvm/Hotspot.scala
```diff
@@ -159,18 +166,19 @@ class Hotspot extends Jvm {
...
+    def asSafepointBean(x: AnyRef): SafepointBean = x match {
+
+      case h: sun.management.HotspotRuntimeMBean => new SafepointBean() {
+        override def getSafepointCount: Long = h.getSafepointCount
+        override def getSafepointSyncTime: Long = h.getSafepointSyncTime
+        override def getTotalSafepointTime: Long = h.getTotalSafepointTime
+      }
+
+      case a => a.asInstanceOf[SafepointBean]
     }
```
sjoerd-vogel, line 180: This cast to an anonymous trait relies on reflection at the scala level and throws a similar warning. Getting around this here by making the trait explict and implementing it by delegation.

ryanoneill:
Hi @sjoerd-vogel! Thank you very much for submitting this pull request for util. Unfortunately, it's not going to be something that we can accept at the moment, because we are still using JDK/JVM 8 at Twitter. Hopefully that is resolved early next year. Until then, it means that we cannot call '.getVMManagement' directly. 

Would it resolve the warnings you are seeing if the code was changed based on version to conditionally call '.getVMManagement' via reflection?

sjoerd-vogel:
> Hi @sjoerd-vogel! Thank you very much for submitting this pull request for util. Unfortunately, it's not going to be something that we can accept at the moment, because we are still using JDK/JVM 8 at Twitter. Hopefully that is resolved early next year. Until then, it means that we cannot call '.getVMManagement' directly.
> 
> Would it resolve the warnings you are seeing if the code was changed based on version to conditionally call '.getVMManagement' via reflection?

I dont think so. AFAIK you get this warning for any reflection on certain classes. Which is why the cast to the anonymous trait was problematic.

If there is anything I can do to speed up the migration of this module to jdk11 I would be happy to help. AFAIK you only need this module to compile against jdk11, I dont think you need to involve the other projects for this and you _should_ still be backwards compatible for running this on older JVMs.

ryanoneill:
> If there is anything I can do to speed up the migration of this module to jdk11 I would be happy to help. AFAIK you only need this module to compile against jdk11, I dont think you need to involve the other projects for this and you should still be backwards compatible for running this on older JVMs.

For our open source code such as this (Util, Finagle, Finatra, TwitterServer, Scrooge), it needs to work both internally and externally. Internally we have a monorepo with source dependencies, and so the cost *must* work with JDK 8 while the company is running JDK 8. So until we've migrated to JDK11 internally, which should happen sometime early next year, we don't have the ability to use any JDK code that does not compile with JDK 8. 

Thank you for the offer to help with this. Unfortunately, my current understanding is that the vast majority of issues are in internal Twitter code.

CEikermann:
@ryanoneill can you give us an update when this pull request can be merged?

Issue #274, opened 2020-04-24
Path to Scala 3


felixbr:
This is an issue for discussing and tracking necessary step to prepare for a Dotty/Scala3 cross-build.

Inspired by the awesome "Scala Love" conference, I attempted to cross-build `util-core` with Dotty `0.23.0-RC1`. So far it went smoother than expected. In the end there were about 4 compile-errors that I couldn't directly address.

In any case there's some work that has to be done before a real cross-build can be achieved. So far I've identified the following but I will try and update this list as we learn more.

**Preparations:**

- [ ] Update sbt to recent version (e.g. `1.3.10`; needed by `sbt-dotty`)
- [ ] Update Scalatest to `3.1.1` (which is cross-published for Dotty)

**Resources for Dotty migration/cross-building:**

- https://github.com/scalacenter/scala-3-migration-guide
- https://github.com/lampepfl/dotty-example-project#getting-your-project-to-compile-with-dotty

I hope you don't mind this type of issue. This is not meant to push anyone, just as a place to discuss and organize things.

Cheers
~ Felix

mosesn:
@felixbr yep, we're perfectly happy to have those kinds of conversations here.  If it turns into a big effort, we'll probably set up a ticket like https://github.com/twitter/finagle/issues/771 for doing it.  We're going to upgrade scalatest soon to support https://github.com/twitter/finatra/pull/527.

I can file an internal ticket around upgrading our sbt version, but we're going through some turbulence with our builds, so there's a good chance that we'll have to hold off on that until we've worked out some of the details.

mosesn:
Oh, also, as we find things that are challenging to migrate, that's probably good feedback for the scala compiler team.  Especially if there are things that scalafix handles poorly.  Please let us know what your experience in trying it out is like!  Thanks for taking this on, I'm excited to see what you find =)

cacoco:
@felixbr FYI, we've updated to sbt 1.3.10 in 6ee7b49a7aa22462648dd9be27e2c577b3a9f003. Scalatest upgrade may take a bit longer. Thanks.

mosesn:
@felixbr sorry for the delay, scalatest was a beast.  It's finally updated https://github.com/twitter/util/commit/db63a6dcf2113606f27e04d6860ad2d0ca72ab5e

felixbr:
Oh, that's great news! :)

mosesn:
[...]

PR #287, opened 2021-05-14
Upgrade to scala-parser-combinators 2.1.1

xerial:
## Problem

Use scala-parser-combinator ~2.0.0~ 2.1.1, which is compatible for Scala 2.11 - Scala 3.0.0.

If a project using twitter-util has a dependency to scala-parser-combinator 2.0.0 with sbt 1.5.x, an error like this will be shown:
```
[info] [error] java.lang.RuntimeException: found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[info] [error]
[info] [error] 	* org.scala-lang.modules:scala-parser-combinators_2.12:2.0.0 (early-semver) is selected over 1.1.2
[info] [error] 	    +- org.wvlet.airframe:airframe-launcher_2.12:21.4.1-38-0bf13a68-20210513-1748-SNAPSHOT (depends on 2.0.0)
[info] [error] 	    +- org.wvlet.airframe:airframe-control_2.12:21.4.1-38-0bf13a68-20210513-1748-SNAPSHOT (depends on 2.0.0)
[info] [error] 	    +- com.twitter:util-core_2.12:21.4.0                  (depends on 1.1.2)
```
Related article: https://www.scala-lang.org/blog/2021/02/16/preventing-version-conflicts-with-versionscheme.html

## Solution

Upgrade the scala-parser-combinator version of twitter/util or adding this setting to build.sbt:
```scala
ThisBuild / libraryDependencySchemes += "org.scala-lang.modules" %% "scala-parser-combinators" % "always"
```

## Result

There should be no behavior change with this upgrade.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -408,6 +408,7 @@ lazy val utilCore = Project(
     libraryDependencies ++= Seq(
       caffeineLib % "test",
       scalacheckLib,
+      "org.scala-lang.modules" %% "scala-parser-combinators" % "2.1.1",
       "org.mockito" % "mockito-core" % mockitoVersion % "test",
     ) ++ {
       if (scalaVersion.value.startsWith("2")) {


yufangong:
Hi @xerial , thanks for the PR! We usually keep all our repositories at the same version of 3rdparty dependencies, and during the tests, we found some issues in other repositories (Scrooge and Finatra), so this would take longer than expected.

felixbr:
Just wanted to mention that this update is also necessary for the Scala3 migration in general (see #290 where I had to include it as well). 🙂

pjfanning:
@xerial would you consider rebasing this to fix the conflict reported on this PR? There has been some progress on https://github.com/twitter/util/issues/274

xerial:
ok. I'll rebase this PR onto develop

xerial:
It's been already rebased. @pjfanning

PR #289, opened 2021-05-24
Update Scala 2.13.1 to 2.13.6

felixbr:
Problem

The latest Scala 2.13.x version is compatible with Scala 3.0.0 and includes many useful compiler-tools (e.g. warnings) to make the transition - and development in general - easier. Even if we don't have a real Scala 3 cross-build, 2.13.6 can be consumed by Scala 3 projects.

Solution

Updating Scala 2.13.1 to 2.13.6

Result

The build uses Scala 2.13.6. I also used the 2.13.x version as default as it has many extremely useful warnings that 2.12.x doesn't have. The only part that needed fixing is some reflection code as 2.13.5 changed encoding of method-parameter names.

As usual, I cannot tell if this clashes with any Twitter-internal things, so please tell me if it does.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -78,8 +78,8 @@ def jdk11GcJavaOptions: Seq[String] = {
 }
 
 val defaultProjectSettings = Seq(
-  scalaVersion := "2.12.12",
-  crossScalaVersions := Seq("2.12.12", "2.13.1")
+  scalaVersion := "2.13.6",
+  crossScalaVersions := Seq("2.12.12", "2.13.6")
 )
 
 val baseSettings = Seq(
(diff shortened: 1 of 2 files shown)


mosesn:
Great!  Again, I don't think this will clash with anything, but let me double check.

mosesn:
Sorry for the delay, this was [merged](https://github.com/twitter/util/commit/7be3a9748a288e9fe3ec2f16d6716c1bf07975b4) in a couple of days ago but I screwed up and merged in a [bug](https://github.com/twitter/util/commit/d22f657cad2db0b663f3155496bb3ce671256fdd) with it too.  Thanks again for the patch, and sorry again about the delay.

PR #291, opened 2021-06-17
Update scalacheck to 1.15.4

felixbr:
**Note**: This is split out from the big WIP PR #290 and can be done separately as preparation for Scala 3.

**Problem**

Scala 3 requires Scalacheck 1.15.x but the build currently uses 1.14.x

**Solution**

Upgrading Scalacheck should be fairly simple as I've yet to see any binary compatibility issues between 1.14.x and 1.15.x.
If we can use the newer version across all Scala versions it will simplify Scala 3 cross-building (otherwise we'd have to use different versions which can be finicky).

**Result**

The tests use Scalacheck 1.15.4. 
We still use `scalatestplus scalacheck-1-14` because there is no `1-15` version available for Scalatest 3.1.x but in my experience there are no binary compatibility issues, so it's worth a try. Since this only affects tests the small risk of a crash is acceptable (and problems should be found in CI).

Cheers
~ Felix

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -23,7 +23,7 @@ val zkDependency = "org.apache.zookeeper" % "zookeeper" % zkVersion excludeAll (
 val guavaLib = "com.google.guava" % "guava" % "25.1-jre"
 val caffeineLib = "com.github.ben-manes.caffeine" % "caffeine" % "2.8.5"
 val jsr305Lib = "com.google.code.findbugs" % "jsr305" % "2.0.1"
-val scalacheckLib = "org.scalacheck" %% "scalacheck" % "1.14.3" % "test"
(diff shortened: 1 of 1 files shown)


smarter:
> Upgrading Scalacheck should be fairly simple as I've yet to see any binary compatibility issues between 1.14.x and 1.15.x.

Scalacheck 1.14 and 1.15 are officially binary-compatible yes: https://github.com/typelevel/scalacheck/releases/tag/1.15.0

tigerlily-he:
Hi @felixbr, Thank you for the PR. I'm currently working on testing the upgrade internally since Twitter works in a monorepo.

tigerlily-he:
Thanks again for the PR. This was merged yesterday. I bumped the Scalacheck version for the other libraries as well.

PR #292, opened 2021-06-17
Match Scala version for 2.13.6 in CI

felixbr:
When I opened the PR to update Scala to 2.13.6 I forgot to update the version for Github Actions. This PR fixes that.

As a side-note: Have you considered using [sbt-github-actions](https://github.com/djspiewak/sbt-github-actions)? It's pretty good in keeping these things consistent and many projects use it successfully.

Cheers
~ Felix

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -20,7 +20,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        scala: [2.12.12, 2.13.1]
+        scala: [2.12.12, 2.13.6]
         java: ['1.8', '1.11']
     runs-on: ubuntu-latest
     steps:


mosesn (approved):

cacoco:
@felixbr FWIW, we have looked at [sbt-github-actions](https://github.com/djspiewak/sbt-github-actions) and well, we have 5 repositories (twitter/util, twitter/scrooge, twitter/finagle, twitter/twitter-server and twitter/finatra) that are developed internally in a monorepo and which we try to keep in sync with each other once in GitHub and that makes off the shelf tools like this a bit challenging to integrate. Thanks for the suggestion. We will probably circle back to it and take another look in the future.

roanta:
Thanks @felixbr, this landed in https://github.com/twitter/util/commit/465601a701d3f604f73e3efcb194bc448efcf492

PR #295, opened 2021-09-09
util/util-logging: Add Scala3 cross-build

felixbr:
This adds Scala3 cross-buliding for the module `util-logging` as discussed in #290

### DIFF
--- a/util-logging/src/main/scala/com/twitter/logging/FileHandler.scala
+++ b/util-logging/src/main/scala/com/twitter/logging/FileHandler.scala
@@ -224,9 +224,10 @@ class FileHandler(
       }
       case Policy.Weekly(weekday) => {
         next.set(Calendar.HOUR_OF_DAY, 0)
-        do {
+        while({ // rewritten from do-while-loop
           next.add(Calendar.DAY_OF_MONTH, 1)
-        } while (next.get(Calendar.DAY_OF_WEEK) != weekday)
+          next.get(Calendar.DAY_OF_WEEK) != weekday
+        }) ()
         Some(next)
       }
     }
--- a/util-logging/src/main/scala/com/twitter/logging/FileHandler.scala
+++ b/util-logging/src/main/scala/com/twitter/logging/FileHandler.scala
@@ -224,10 +224,12 @@ class FileHandler(
       }
       case Policy.Weekly(weekday) => {
         next.set(Calendar.HOUR_OF_DAY, 0)
-        while({ // rewritten from do-while-loop
+
+        next.add(Calendar.DAY_OF_MONTH, 1)
+        while (next.get(Calendar.DAY_OF_WEEK) != weekday) {
           next.add(Calendar.DAY_OF_MONTH, 1)
-          next.get(Calendar.DAY_OF_WEEK) != weekday
-        }) ()
+        }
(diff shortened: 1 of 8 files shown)


util-logging/src/test/scala/com/twitter/logging/FileHandlerTest.scala
```diff
@@ -285,7 +285,7 @@ class FileHandlerTest extends AnyWordSpec with TempFolder {
 
         val handler = FileHandler(
           filename = folderName + "/test.log",
-          rollPolicy = Policy.MaxSize(maxSize.bytes),
+          rollPolicy = Policy.MaxSize(maxSize.toLong.bytes),
```
felixbr: This conversion is necessary in Scala3 because the extension method is only defined on `Long`, and it no longer implicitly converts to `Int`. In #290 I added an implicit conversion as forwarder but that's not landed in `develop` yet.

felixbr:
Not sure why `com.twitter.app.AppTest` is failing (different module and works locally), might be a flaky test.

util-logging/src/main/scala/com/twitter/logging/FileHandler.scala
```diff
@@ -224,9 +224,10 @@ class FileHandler(
       }
       case Policy.Weekly(weekday) => {
         next.set(Calendar.HOUR_OF_DAY, 0)
-        do {
+        while({ // rewritten from do-while-loop
```
hamdiallam: not a fan of scala3 rewrite of this. Can we do it ourselves?

```
next.add(Calendar.DAY_OF_MONTH, 1)
while (next.get(Calendar.DAY_OF_WEEK) != weekday) {
   next.add(Calendar.DAY_OF_MONTH, 1)
}
```

util-logging/src/test/scala/com/twitter/logging/FileHandlerTest.scala
hamdiallam: Can we strip out the forwarders into a separate PR?

hamdiallam (approved):

util-logging/src/main/scala/com/twitter/logging/FileHandler.scala
felixbr: Yeah, I didn't like it either but didn't spend more time thinking about it. I'll change it to your suggestion.

mosesn (approved):
looks great, thanks so much @felixbr!

util-logging/src/test/scala/com/twitter/logging/FileHandlerTest.scala
felixbr: I can make a PR based on what I've already added in #290 to keep source compatibility for Scala3

felixbr: The separate PR is here: #296

felixbr:
I rebased the PR onto `develop`, removed the now unnecessary `.toLong` conversion, and fixed one more deprecation warning regarding Scala 3.

I think the PR can now be merged, if you like. 🙂

joybestourous:
Merged! b8a30a7

PR #296, opened 2021-09-23
util/util-core: Fix conversions to units from Int in Scala3

felixbr:
This fixes lost source-compatibility for unit conversions from `Int` values in Scala3. As discussed in #295 I extracted/adapted this from #290 ([previous discussion](https://github.com/twitter/util/pull/290/files#r680089697)).

Previously something like `1.toInt.second` did compile in Scala2 (with default settings) but didn't in Scala3.

I added forwarders for all units which use `Long` as base, as that seemed reasonable and safe. I didn't add an implicit conversion from `Float` to `Double` because floating-point types are dumb and I don't trust the conversion to be lossless in all cases.

I'll adapt #295 once this is merged.

### DIFF
--- a/util-core/src/main/scala/com/twitter/conversions/DurationOps.scala
+++ b/util-core/src/main/scala/com/twitter/conversions/DurationOps.scala
@@ -1,7 +1,9 @@
 package com.twitter.conversions
 
 import com.twitter.util.Duration
+
 import java.util.concurrent.TimeUnit
+import scala.language.implicitConversions
 
 /**
  * Implicits for writing readable [[com.twitter.util.Duration]]s.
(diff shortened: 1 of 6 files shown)


felixbr:
Would it make sense to update Scala from `3.0.2-RC1` to `3.0.2` and add it to CI? This would help with avoiding regressions like this in the future. Currently PRs are not automatically tested against Scala 3.

hamdiallam:
> Would it make sense to update Scala from `3.0.2-RC1` to `3.0.2` and add it to CI? This would help with avoiding regressions like this in the future. Currently PRs are not automatically tested against Scala 3.

Yes, let's update to 3.02! Was waiting for that release which includes a few bug fixes which first pushed us to using RC1. I'm for including it in CI but I wonder if we need the entire repo crossbuilding before that could work?

Now that we have forwarders. If it's not a hassle, is it possible to undo the `toLong` conversions the interns added in the projects that crossbuild scala 3 to originally fix this problem. Otherwise LGTM! I can undo them in a separate pr otherwise

hamdiallam (approved):

felixbr:
> I'm for including it in CI but I wonder if we need the entire repo crossbuilding before that could work?

One way to do it is to have a separate aggregate-module or even just an alias which tests only the modules which currently support Scala3.

> If it's not a hassle, is it possible to undo the toLong conversions the interns added in the projects that crossbuild scala 3 to originally fix this problem.

I did a quick search when I wrote this PR and didn't see any obvious parts where `.toLong` was added for backwards compat but I might have missed something. Once this is merged, I will undo it in my other PR and rebase onto `develop`.

yufangong:
Hey @felixbr, going to merge this and #295 in, is there an order you prefer which one got merged first, looks like we need to adapt them at the end. Thanks!

felixbr:
The idea was to merge this first and rebase the other one, so the workaround there isn't needed anymore.

I can do the rebase and adaptation once this is merged, if you want.

yufangong (approved):

yufangong:
Sorry for the delay, this is merged at https://github.com/twitter/util/commit/ad8b22b9adbe1958f33b08cf133a9baac2c0aa9a, let us know when the #295 is ready. thank you!

PR #299, opened 2022-05-15
Random fixes

davidefalcone21:
replaced substring with dropright for best practice


cacoco:
@Demigod21 can we ask that you remove the `.deepsource.toml` file from this commit? Thanks!

davidefalcone21:
I didn't notice it, now I removed it, thank you and have a nice day :)

Il giorno lun 16 mag 2022 alle ore 19:44 Christopher Coco <
***@***.***> ha scritto:

mosesn:
@Demigod21 sorry for the delay in looking at this!  can you tell me more about the motivation behind this patch?  the nice thing about the current approach is that it only uses methods are native to String, so we don't need to use the StringOps implicit for dropRight.  this makes compilation a tad faster, and can shave some unnecessary work.  What do you think makes dropRight more appropriate here?

mosesn:
Also, would you mind signing the CLA?  Thank you!

davidefalcone21:
Good morning!

I already signed the CLA.
Anyway feel free to discard my PR, I'm a newbie in the Scala world and I'm
trying to improve my skills with the documentation and deepsource and I
understood that dropRight is more efficient than substring.

But as I said, I'm just a newbie, so thank you for your explanation and
have a nice weekend :)

Il giorno mar 31 mag 2022 alle ore 22:04 Moses Nakamura <
***@***.***> ha scritto:

mosesn:
@Demigod21 OK!  Well, welcome to Scala and Open Source :)  We won't accept this PR for now, but we look forward to future PRs from you

PR #304, opened 2022-06-30
Update scala versions

felixbr:
This is primarily to get away from the release candidate version of Scala 3. I also updated the other Scala versions for general maintenance.

edit: I missed PR #300 which already updates Scala 3 to a non-RC version, my apologies. I'll let you be the judge which PR to accept and will modify/rebase/close mine accordingly.

edit2: I wanted to add CI for Scala 3 as well but scoverage isn't quite there yet and some build adjustments are needed, so this will require a separate PR.

Cheers
~ Felix

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -82,8 +82,10 @@ def jdk11GcJavaOptions: Seq[String] = {
   )
 }
 
+// Upgrading to a later version requires also upgrading sbt-scoverage, Scalatest, and
+// some modules like util-mock and util-test.
 val _scalaVersion = "2.13.6"
-val _crossScalaVersions = Seq("2.12.12", "2.13.6")
+val _crossScalaVersions = Seq("2.12.15", "2.13.6")
 
 val defaultScalaSettings = Seq(
   scalaVersion := _scalaVersion,
@@ -91,7 +93,7 @@ val defaultScalaSettings = Seq(
 )
 val defaultScala3EnabledSettings = Seq(
   scalaVersion := _scalaVersion,
-  crossScalaVersions := _crossScalaVersions ++ Seq("3.0.2-RC1")
+  crossScalaVersions := _crossScalaVersions ++ Seq("3.1.3")
 )
 
 // Our dependencies or compiler options may differ for both Scala 2 and 3. We branch here
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,6 +2,6 @@ resolvers += Classpaths.sbtPluginReleases
 resolvers += Resolver.sonatypeRepo("snapshots")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.4.1")
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.9.1")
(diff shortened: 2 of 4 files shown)


felixbr:
The failing test seems to be the same flaky one as in the past

felixbr:
The last commit is a bit off-topic but hopefully it fixes the flaky test which has been bugging me for a while now.

build.sbt
```diff
@@ -82,16 +82,18 @@ def jdk11GcJavaOptions: Seq[String] = {
...
+val _crossScalaVersions = Seq("2.12.15", "2.13.6")
 
 val defaultScalaSettings = Seq(
   scalaVersion := _scalaVersion,
   crossScalaVersions := _crossScalaVersions
 )
 val defaultScala3EnabledSettings = Seq(
   scalaVersion := _scalaVersion,
-  crossScalaVersions := _crossScalaVersions ++ Seq("3.0.2-RC1")
+  crossScalaVersions := _crossScalaVersions ++ Seq("3.1.3")
```
pjfanning, line 96: There is a merge conflict because a PR was merged that changed this to 3.0.2 - building with 3.0.2 is better imho because classes will work with Scala 3.0.x and 3.1.x. If you compile with 3.1.3, then the classes will not work with Scala 3.0.x runtime.

project/plugins.sbt
```diff
@@ -2,6 +2,6 @@ resolvers += Classpaths.sbtPluginReleases
 resolvers += Resolver.sonatypeRepo("snapshots")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-site" % "1.4.1")
-addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.9.1")
+addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.9.3")
```
pjfanning, line 5: sbt-scoverage 2.0.0 is now out and upgrade seems to be smooth (from my experience with other projects)

build.sbt
felixbr: Are you sure about that? I thought binary compatibility was finally solved by the Tasty mechanism. 

The libraries published to Maven also don't specify anything beyond "_3" as the version suffix.

If there is really still a problem I don't mind building with 3.0.x for the moment, though.

pjfanning: Have a read of https://issues.apache.org/jira/browse/LOG4J2-3524 and its link

project/plugins.sbt
felixbr: I tried that first but ran into all sorts of problems, so I changed the PR to be more conservative. My main goal here is using a recent Scala 3 version.

I only update Scala 2.x versions until it became painful to go further (e.g. 2.12.16 and 2.13.8 both caused issues).

build.sbt
felixbr: Ok, I see. That's unfortunate.

I will rework/close my PRs as soon as I find the time.

felixbr:
Almost everything I'm trying to do here has landed in `develop` via other commits, so I'm closing this PR and will rebase #305 and #306 onto `develop` when I have time to do so.

PR #305, opened 2022-06-30
support scala 3 build for util-tunable

felixbr:
This PR builds on #304 so that should be merged before this one imo.

With the new Jackson version supporting Scala 3 directly, `util-tunable` was just a config change.

I'm not sure why `scala-collection-compat_3` was excluded from the Scala 3 settings. It was probably an attempt to solve the "conflicting version suffix" error from sbt. I moved it to the Scala 2 settings but I can also remove it if you prefer.

### DIFF
--- a/util-tunable/src/main/scala/com/twitter/util/tunable/TunableMap.scala
+++ b/util-tunable/src/main/scala/com/twitter/util/tunable/TunableMap.scala
@@ -5,6 +5,7 @@ import java.util.function.BiFunction
 import scala.annotation.varargs
 import scala.collection.mutable
 import scala.jdk.CollectionConverters._
+import scala.reflect.ClassTag
 
 /**
  * A Map that can be used to access [[Tunable]]s using [[TunableMap.Key]]s.
@@ -133,7 +134,7 @@ object TunableMap {
   }
 
   object Key {
-    def apply[T](id: String)(implicit m: Manifest[T]): Key[T] =
+    def apply[T](id: String)(implicit m: ClassTag[T]): Key[T] =
       Key[T](id, m.runtimeClass.asInstanceOf[Class[T]])
   }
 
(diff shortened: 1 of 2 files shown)


felixbr:
CI failed because sbt-scoverage was updated to 2.0.0 [here](https://github.com/twitter/util/commit/8d06d68b477d011d912cb12264913e80c7d8b62b). 

Once `develop` is green again, I can rebase this PR once more, which should fix it.

util-tunable/src/main/scala/com/twitter/util/tunable/TunableMap.scala
```diff
@@ -133,7 +134,7 @@ object TunableMap {
   }
 
   object Key {
-    def apply[T](id: String)(implicit m: Manifest[T]): Key[T] =
+    def apply[T](id: String)(implicit m: ClassTag[T]): Key[T] =
```
felixbr, line 137: `Manifest` no longer exists in Scala 3 and there is no equivalent replacement.

Since it is only used to retrieve the runtime class (which is erased anyway), `ClassTag` should work the same way here, as far as I understand it.

felixbr:
I've rebased this onto `develop`.

I've also had to change the Scala 3 version form `3.0.2` to `3.2.2` because `util-tunable` depends on `util-jackson`, which is not yet ported for Scala 3 natively. 
`util-jackson` in turn depends on `jackson-module-scala` which was compiled with Scala `3.2.x` already, so we either have to port `util-jackson` to support Scala 3 natively (which currently is blocked on `json4s`) or we have to also use Scala `3.2.x`.

Given that we currently don't have CI-checks for Scala 3 (see #310 for that) and it's still somewhat of a moving target, I'd rather just go for the up-to-date Scala 3 version, which we want anyway in the long run.

edit: I should add that I ran `++3.2.2 Test/compile` in `util` to make sure I didn't break any other modules which were already ported to Scala 3.

PR #307, opened 2022-07-27
See what happens

vkostyukov:


felixbr:
Hi,

I noticed you're hitting similar trouble with `sbt-scoverage` as I have recently. Maybe it would be helpful to run coverage only for the main Scala version (i.e. 2.13.6 currently) instead of all versions.

The code and tests are the same either way, so imo it would be sufficient to only do it for the main version and it might make the `sbt-scoverage` version stuff easier.

You can take a look at #308 as an example where I've excluded coverage for Scala 3 and it seems to work nicely.

Cheers
~ Felix

edit: There is also a `coverageEnabled` setting in sbt, which might be helpful but I didn't have too much luck with it, so I added conditionals in the CI-build instead.

bryce-anderson:
Looks like it worked.

PR #308, opened 2022-07-28
build: Add Scala 3 to CI matrix

felixbr:
Several people have worked on Scala 3 support of various modules, which is great.

Scala 3 support currently isn't part of CI, which means it is only validated locally by the person currently working on it. This causes already working cross-builds to break by accident (e.g. when updating versions).

To avoid this and make working on Scala 3 more efficient I'm trying to setup at least basic CI for it.

Coverage is currently not really working with Scala 3.0.2, so I'm trying to ignore it for Scala 3 only.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -21,7 +21,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        scala: [2.12.12, 2.13.6]
+        scala: [2.12.12, 2.13.6, 3.0.2]
         java: ['1.8', '1.11']
     runs-on: ubuntu-latest
     steps:
@@ -52,4 +52,4 @@ jobs:
             find ~/.sbt         -name "*.lock"               -delete
           fi
       - name: test
-        run: ${{ format('./sbt ++{0} clean test', matrix.scala) }}
+        run: ${{ format('./sbt "++{0} -v clean; ++{0} test"', matrix.scala) }}


felixbr:
Ok, so apart from the usual flaky test in `util-app` everything seems to be working decently.

I've downgraded sbt-scoverage to `1.9.1` though, because `develop` is currently not green.

I'm rebasing this onto develop to make it ready for merging, even if it means the CI will be red on this PR in the meantime. 🙂

bryce-anderson:
@felixbr, do you mind rebasing on master to get the scoverage fix pulled in and presumably turn the CI checks green?

felixbr:
@bryce-anderson There you go. 

The one failing test is the usual flaky one in util-app. I'd recommend setting its timeout much higher because it's really annoying that PRs are red all the time because of it.

It's waiting for an interrupt but CI is too slow so instead it throws a timeout:

```
[info] - App: exit functions properly capture mix of non-fatal and fatal exceptions *** FAILED ***
[info]   class com.twitter.util.TimeoutException did not equal class java.lang.InterruptedException (AppTest.scala:418)
```

felixbr:
I rebased the PR onto `develop`.

felixbr:
I rebased the PR onto `develop`.

The two failed runs are the usual flaky test in `util-app` (see #310)

rbscgh:
can we bump this to a LTS version of 3 like 3.3.4 ?

felixbr:
@rbscgh We could, but I think it's fair to say Twitter (or whatever is left of it) has given up on open source contributions.

Until I see a clear indication to the contrary I won't waste more time on this.

That said, if there's a community fork let me know and I'll gladly help out.

PR #312, opened 2023-03-20
Expose underlying slf4j Logger

shappley:
Updates `util.Logger` to expose the underlying `slf4j.Logger` instance. This is beneficial for those with projects with several SLF4J Loggers, and follows the pattern in the [scala-logging project](https://github.com/lightbend-labs/scala-logging/blob/main/src/main/scala/com/typesafe/scalalogging/Logger.scala#L82).


cacoco:
Based on the link to the scala-logging project, isn't that the same as [here](https://github.com/twitter/util/blob/1904b60ab81c07e1257f30dca9d41494899148a9/util-slf4j-api/src/main/scala/com/twitter/util/logging/Logger.scala#L69)?

